PR #114, opened 2020-07-16
Whitelist_PR2

StefanVriend:
Second pull request for whitelist branch.


R/approved_list.R
```diff
@@ -28,13 +28,20 @@ create_approved_list <- function(new_approved_list){
...
+
     Individual_approved_list <- dplyr::bind_rows(approved_list$Individual_approved_list,
-                                                 new_approved_list$Individual_approved_list)
+                                                 new_approved_list$Individual_approved_list) %>%
+      tidyr::drop_na()
+
     Location_approved_list <- dplyr::bind_rows(approved_list$Location_approved_list,
-                                               new_approved_list$Location_approved_list)
+                                               new_approved_list$Location_approved_list) %>%
+      tidyr::drop_na()
```
LiamDBailey: Do we expect there to be NAs? If we use `drop_na` with no column specified it will drop any row where there is at least one NA, so we should use it with caution (or specify the columns where we NAs are unacceptable).

R/quality_check.R
```diff
@@ -236,6 +254,52 @@ quality_check <- function(R_data,
...
+    '\\newpage',
+    '# Verified records',
+    '',
+    '## Brood data',
+    '',
+    '```{r, echo=FALSE}',
+    '    purrr::pwalk(.l = Brood_approved_list,
+                 .f = ~{ cat(paste0("Record with BroodID ", ..2,
+                                    " and PopID ", ..1,
+                                    " has been verified by the data owner and does no longer violate check ",
```
LiamDBailey, line 266: In theory it still violates the check...it's just that it doesn't matter anymore 😄 

```suggestion
                                    " has been verified by the data owner and is trustworthy",
```

R/quality_check.R
```diff
@@ -236,6 +254,52 @@ quality_check <- function(R_data,
...
+                  })',
+    '```',
+    '',
+    '## Capture data',
+    '',
+    '```{r, echo=FALSE}',
+    '    purrr::pwalk(.l = Capture_approved_list,
+                 .f = ~{ cat(paste0("Record with CaptureID ", ..2,
+                                    " and PopID ", ..1,
+                                    " has been verified by the data owner and does no longer violate check ",
```
LiamDBailey, line 277:
```suggestion
                                    " has been verified by the data owner and is trustworthy",
```

R/quality_check.R
```diff
@@ -236,6 +254,52 @@ quality_check <- function(R_data,
...
+                  })',
+    '```',
+    '',
+    '## Individual data',
+    '',
+    '```{r, echo=FALSE}',
+    '    purrr::pwalk(.l = Individual_approved_list,
+                 .f = ~{ cat(paste0("Record with IndvID ", ..2,
+                                    " and PopID ", ..1,
+                                    " has been verified by the data owner and does no longer violate check ",
```
LiamDBailey, line 288:
```suggestion
                                    " has been verified by the data owner and is trustworthy",
```

R/quality_check.R
```diff
@@ -236,6 +254,52 @@ quality_check <- function(R_data,
...
+                  })',
+    '```',
+    '',
+    '## Location data',
+    '',
+    '```{r, echo=FALSE}',
+    '    purrr::pwalk(.l = Location_approved_list,
+                 .f = ~{ cat(paste0("Record with LocationID ", ..2,
+                                    " and PopID ", ..1,
+                                    " has been verified by the data owner and does no longer violate check ",
```
LiamDBailey, line 299:
```suggestion
                                    " has been verified by the data owner and is trustworthy",
```

R/approved_list.R
StefanVriend: This is added because the dummy_approved_list contains full-NA rows. Do you see an alternative way to deal with that?

R/quality_check.R
StefanVriend: Sharp!

I still want to include the CheckID there, so what do you think of the following:
"Record with BroodID X and PopID Y violates check (CheckID Z) but has been verified by the data owner and is trustworthy."

R/approved_list.R
StefanVriend: I've tested this again and I was wrong: it is not needed at all. So I removed the ```drop_na```.

Issue #40, opened 2019-09-17
Check that birds do not have multiple active nests


LiamDBailey:
We can test this in two ways:

1. If an individual has a nest with a known laying date and known hatch/fledge date, we know that this individual could not have any other active nests in this period. This should be an error.

2. If an individual has a nest with a known laying date and a known brood/fledge number, we know that this individual should not have any other active nests in the period:

> Laying Date + Minimum Incubation Time + Minimum Fledge Time

This should be a warning, as we are making assumptions about the length of incubation and chick development.

**N.B.** In cases where no laying date is known or no hatch/fledge information is available, we cannot test this.

Issue #39, opened 2019-09-17
Check that coordinates of all locations are expected


LiamDBailey:
In the meta-data of each population we have a coordinate for the study site (WSG 84). Coordinate locations of all capture locations should be close to these centre points.

StefanVriend:
As suggested by @zuzulaz: 
These packages may be useful for checking the format of coordinates

CoordinateCleaner
https://github.com/ropensci/CoordinateCleaner

biogeo: Point Data Quality Assessment and Coordinate Conversion
https://cran.r-project.org/web/packages/biogeo/index.html

PR #115, opened 2020-08-14
New quality checks

StefanVriend:


StefanVriend:
Remaining tasks:

- Check unusual coordinates (see #39). Cluster the lat/long and throw warnings for cases outside of these clusters. Only warnings, because there may be unusual records (e.g. dead bird found outside of the study site).
- Check overlapping nests (see #40). Can use lubridate ‘overlap’ function.
- Update documentation (see #116).
- Change reference value checks to use data rather than arbitrary values (see #117).

StefanVriend:
Reference values in Brood_data (see check_values_brood, from line 716 onwards) are now generated by the data.

If number of observations for a particular population-species combination is too low, I suggest to skip these records in the check rather than comparing against Hoge Veluwe/NIOO data. This comparison might be more or less valid for variables like clutch size, but is far from reasonable in the case of phenology-related variables, like laying date.

Reference value determination:

- Values for ClutchSize, BroodSize and NumberFledged are considered a warning if larger than the 99th percentile, and an error if negative or 4 times the 99th percentile. 

- Values for LayDate are considered a warning if earlier than the 1st percentile or later than the 99th percentile, and an error if earlier than Jan 1st or later than Dec 31st.

LiamDBailey:
@StefanVriend not a bad idea. I think it might be better to hold off on assigning errors/warnings rather than giving poor errors/warnings that we'll then have to discuss with the data owners. Do you know how many populations currently would **_not_** be tested using this 50 observation cutoff?

R/brood_check.R
```diff
@@ -68,15 +75,15 @@ brood_check <- function(Brood_data, Individual_data, check_format=TRUE){
 
   check_list[3,3:4] <- compare_brood_fledglings_output$CheckList
 
-  # - Compare laying and hatching dates
-  message("B4: Comparing laying and hatching dates...")
+  # - Compare lay and hatchdates
```
LiamDBailey, line 78:
```suggestion
  # - Compare lay and hatch dates
```

R/create_reference_values.R
```diff
@@ -14,81 +15,110 @@
 #'
 #' @export
 
-create_reference_values <- function() {
+##FIXME: remove this function if check_values_capture is updated.
```
LiamDBailey, line 18: Can now be removed?

StefanVriend:
These population-species combinations do not make the cut:
```
   Species PopID     n
   <chr>   <chr> <int>
 1 POEPAL  BUU       1
 2 SITEUR  BUU      49
 3 POEPAL  HOG      27
 4 PERATE  LIE       1
 5 PERATE  MIS      10
 6 PERATE  MUR       1
 7 FICHYP  OOS      10
 8 PASMON  OOS       2
 9 POEPAL  OOS       1
10 SITEUR  OOS      15
11 PERATE  VLI      25
12 PERATE  WAR      40
13 SITEUR  WAR      10
14 POEPAL  WES       7
```

LiamDBailey (commented):
Looks good, just a few minor suggestions. I think the idea to only use local data when sample size is large enough is fine, in which case we delete the code in `create_reference_values.R` (l.17+)

R/create_reference_values.R
StefanVriend: The first part of that function, so the lines concerning brood data, can be removed, but the lines concerning capture data have to stay because the check_values_capture function has not been updated yet.

StefanVriend:
Reference values in Capture_data (see `check_values_capture`, from line 245 onwards) are now generated by the data (see #117).

For adult mass, adult tarsus and chick tarsus records, we use the same threshold as with brood values: > 50 observations, reference values are created; < 50 check is skipped.

For chick mass, the function tries to fit a logistic growth model (for which I adjusted the code that @LiamDBailey made, `calculate_chick_mass_cutoffs`). If this fails, the function will continue and create reference values per age. Ages with < 50 observations are skipped.

Currently the threshold is set at 50 observations. However, we might miss an obvious outlier if there are 49 "correct" records and 1 "incorrect" record, if that is an obvious outlier (as it pulls the 99th percentile up). I consider raising this threshold to 100 instead. 

What do you think? @LiamDBailey, @ChloeRN, @zuzulaz

Once this PR is ready to be merged, I will check the conflicting files. I suspect the conflict with the internal data update PR that was merged a week ago.

R/brood_check.R
```diff
@@ -757,10 +757,10 @@ check_values_brood <- function(Brood_data, var) {
   } else if(var %in% c("LayDate", "LayDate_observed")) {
 
     ref <- Brood_data %>%
-      dplyr::filter(!is.na(!!rlang::sym(var))) %>%
+      dplyr::filter(!is.na(!!rlang::sym(var)) & !is.na(Species)) %>%
       dplyr::group_by(BreedingSeason) %>%
-      # Transform dates to julian days to calculate quantiles
-      dplyr::mutate(!!paste0(var, "_julian") := lubridate::yday(!!rlang::sym(var))) %>%
+      # Transform dates to Julian days (while accounting for year) to calculate quantiles
+      dplyr::mutate(!!paste0(var, "_julian") := lubridate::yday(!!rlang::sym(var)) + (lubridate::year(!!rlang::sym(var)) - BreedingSeason) * 365) %>%
```
LiamDBailey, line 763: Perhaps an easier approach would be to just find the difference between dates. That way you don't have to multiply by 365 (which could cause some weirdness with leapyears). Suggestion below, not much prettier but I think a bit more robust.

```suggestion
      dplyr::mutate(!!paste0(var, "_julian") := as.numeric(!!rlang::sym(var) - lubridate::ymd(paste(BreedingSeason, "1", "1", sep = "-")) + 1)) %>%
```

R/brood_check.R
```diff
@@ -883,8 +883,9 @@ check_values_brood <- function(Brood_data, var) {
 
                                Brood_data %>%
                                  dplyr::group_by(BreedingSeason) %>%
-                                 # Transform dates to julian days to compare to julian day reference values
-                                 dplyr::mutate(!!paste0(var, "_julian") := lubridate::yday(!!rlang::sym(var))) %>%
+                                 # Transform dates to Julian days (while accounting for year)
+                                 # to compare to Julian day reference values
+                                 dplyr::mutate(!!paste0(var, "_julian") := lubridate::yday(!!rlang::sym(var)) + (lubridate::year(!!rlang::sym(var)) - BreedingSeason) * 365) %>%
```
LiamDBailey, line 888: As above

R/brood_check.R
```diff
@@ -938,8 +942,9 @@ check_values_brood <- function(Brood_data, var) {
...
```
LiamDBailey, line 947: As above

LiamDBailey (changes requested):
Changes look fine but it's currently failing the tests (running `devtools::test()`). It fails at setup-xyz.R. Try and get this working before we merge.

StefanVriend:
Thanks for the review. I'm trying to run the test, but there's an issue running the pipelines for the test. In line 115, when cycling through the directories something goes wrong. I have no clue why it fails as I have not changed anything in `run_pipelines`...

PR #129, opened 2020-10-20
PFN pipeline corrections

ChloeRN:
This update contains some adjustments to maximise the use of information to determine breeder age. 

After this update, age is calculated from capture histories BEFORE subsetting the ringing data to contain only EDM sites. That way, age calculation can take into account individuals ringed in other areas (immigrants). Furthermore, minimum parent age in the brood data is now treated as "true" (not minimum) age if the bird is classified as a local recruit in the raw data (as discussed with Malcolm).

Furthermore, after realising that "brood" raw data contained more detailed information on parent age than the ringing raw database, I made some adjustments to the pipeline so that additional age data is now merged from brood raw data into capture data.


ChloeRN:
@LiamDBailey these are the updates I mentioned during the last Developer's meeting. Details are in the pull request message. Everything works smoothly for me even after these changes, but may be worth double-checking quickly before the merge.

LiamDBailey (commented):
@ChloeRN sorry for the delay. This all looks fine. One question, I noticed that we now have 'format_PFN' but e.g. `create_brood_EDM`. When you include the other pipelines like Nagshead will these have there own `create_brood_NAG` function within the `format_PFN`?

ChloeRN:
@LiamDBailey thanks for the feedback. 
I have been working intensely on dummy pipelines for 6 other datasets from Malcolm (at least 5 of which will be submitted to SPI-birds in the near future), and have been gradually expanding and generalising the "original" (EDM) pipeline.  

I am actually about to branch off again now to write a PFN "master pipeline" based on this, and from the way it looks, I will be changing the function names to "PFN", as the function will run on data from several populations simultaneously.
So once Malcolm submits more data, the current format_PFN will be replaced with the master pipeline format_PFN, which will contain subfunctions also ending in "PFN".

LiamDBailey:
Good idea. I think it's always good to have data stored together like this if possible because it reduces the risk of having different versions for different pops from the same data owner.

PR #133, opened 2020-10-26
Subsetting function

ChloeRN:
I made the following updates to the previously existing subset_datarqst() function:
- Several fixes to ensure it runs correctly
- Functionality extended to work on Mac
- Added functionality to choose whether or not conflicted species individuals are included

Further, I made some basic tests for the function. Suggestions on what else to add to the tests are very welcome. 

The pull request also contains some changes to make run_pipelines() work smoothly on Mac. Specifically, I changed the naming of paths, and added functionality to skip pipelines for populations with Access raw data when OS = Mac.


ChloeRN:
It would be good to test the function on the "complete" set of standard data. I was only able to test it on standard data for which pipelines currently run on Mac. 
This means the following were not included in my workflow/tests:
- NIOO and AMM (Access databases)
- HAR and KEV (Paradox databases. Not currently working on Mac due to issue with python libraries)
- MON (throws several warnings, plus errors for populations MON & MTV)

(In the tests, I further dropped BOS, PEE, and WYT out of purely practical reasons, i.e. they take very long to run).

R/subset_datarqst.R
```diff
@@ -37,66 +45,141 @@
...
+  }else{
+    standard_data <- test_data
+  }
 
-  standard_data <- readRDS(file = file)
 
   if(!is.null(filter)){
 
-    unique_pops <- unique(stringr::str_split(string = filter, pattern = "_")[[1]])
+    unique_pops <- unique(matrix(unlist(stringr::str_split(string = filter, pattern = "_")), ncol = 2, byrow = T)[,1])
```
StefanVriend, line 64: I found it slightly difficult to read this line; however, when it works it works :) 

`stringr::str_split` has an argument "simplify" that returns a matrix rather than a list, maybe that helps the readability a bit (so for instance `unique(stringr::str_split(string = filter, pattern = "_", simplify = TRUE)[,1])`).

R/subset_datarqst.R
```diff
@@ -37,66 +45,141 @@
...
-                            filter = NULL){
+                            filter = NULL,
+                            include_conflicting = FALSE,
+                            output_type = "R",
+                            return_R = FALSE,
+                            save = TRUE,
+                            test = FALSE){
+
+  if(!test){
+    readRDS(file = file)
```
StefanVriend, line 56: Should this say the following?
```suggestion
    standard_data <- readRDS(file = file)
```

StefanVriend (changes requested):
Great work! I had one error (see my suggested change), but after that, the code ran fine.
I did the following subset `subset_datarqst(PopID = "HOG")` and it took quite a long while to run (I guess 10 minutes). I'm not sure what part of the code took so long. It would be nice if a few progress bars are added so we can track the pace of the subsetting function :)

ChloeRN (commented):
This now works for me on MacOS. 

@zuzulaz it would be great if you could check if it works for you too. I left some notes on how to set up the python environment and libraries on MacOS under issue #134 )

@LiamDBailey and @StefanVriend, would also be good to know if it still works on windows too.

R/run_pipelines.R
```diff
@@ -65,6 +65,35 @@ run_pipelines <- function(path = choose_directory(),
...
+                    ". To obtain standard format data for these populations, please run on a Windows OS.")
+              )
+    }
+
+    PopID <- PopID[which(!(PopID%in%PopID_Access))]
+
+  }else if(grepl(pattern = 'windows', x = OS)){
+
+    PopID <- PopID
+
```
LiamDBailey, line 88: This is redundant right? PopID is already defined before the ifelse, so no need to redefine it here.
```suggestion
```

R/run_pipelines.R
```diff
@@ -65,6 +65,35 @@ run_pipelines <- function(path = choose_directory(),
...
+  }else{
+    stop(paste0('Operating system ', OS, ' not supported'))
+  }
+
+  if(length(PopID) == 0){
+    stop(paste0('None of the selected pipeline(s) could not be run due to OS incopatibility. Please run on a Windows OS.'))
```
LiamDBailey, line 94:
```suggestion
    stop(paste0('None of the selected pipeline(s) could not be run due to OS incompatibility. Please run on a Windows OS.'))
```

R/subset_datarqst.R
```diff
@@ -15,9 +15,17 @@
 #' @param Species Character vector of six letter species codes. Include all species that are requested.
 #' @param filter Character vector of unique population species combinations (in the format PopID_Species).
 #' Include all unique population species combinations requested.
-#'
-#' @return Create 4 .csv files at the location of the standard data. Files will be given a unique number
-#' based on the date when the subset was performed.
+#' @param include_conflicting is FALSE. If include_conflicting = TRUE, individuals of conflicting species which
+#' were identified as a species contained in `Species` at least once are included in the subset of data. If
+#' include_conflicting = FALSE, these individuals are removed.
```
LiamDBailey, line 20: A think we can simplify the help documentation here a bit.
```suggestion
#' @param If include_conflicting = TRUE, individuals with conflicting species information
#' are included in the subset of data. If
#' include_conflicting = FALSE (default), these individuals are removed.
```

R/subset_datarqst.R
```diff
@@ -15,9 +15,17 @@
...
+#' @param output_type is 'R' and can be set to 'csv'. If output_type is 'csv' 4 .csv files will be created in the save path.
+#' If output_type is 'R' an .RDS file will be created in the save path, and an R object in the
+#' running R session if return_R = TRUE.
+#' @param return_R is FALSE. If return R = TRUE, the subset of data is also saved to the current
+#' R Session as an R object (list).
+#' @param save is TRUE. If save = FALSE, the subset of data is not saved into the save path.
```
LiamDBailey, line 26: Do we need all these save options? The idea of the function is to create a subset of the standard format so I would think that we would *always* want to save to path. Thoughts?

R/subset_datarqst.R
```diff
@@ -15,9 +15,17 @@
 #' @param Species Character vector of six letter species codes. Include all species that are requested.
 #' @param filter Character vector of unique population species combinations (in the format PopID_Species).
```
LiamDBailey, line 16: I guess I created these, but I think we should specify to the user that they're alternative arguments (i.e. you can either provide PopIDSpecies combos or provide species and pops separately)

R/subset_datarqst.R
```diff
@@ -37,66 +45,141 @@
...
+      dplyr::filter(.data$PopID %in% {{PopID}})
 
   }
 
-  message("Output subset as .csv files...")
+  #If keeping conflicted species: remove IndvIDs that do not appear in capture data from individual data
+  #(this ensures that only data on individuals identified at least once as one of the species of interest are retained)
+  if(include_conflicting){
+    output_individual <- output_individual %>%
+      dplyr::filter(!(.data$Species %in% c('CONFLICTED', 'CCCCCC') & !(.data$IndvID %in% output_capture$IndvID)))
```
LiamDBailey, line 123: Not sure I understand this filter. Wouldn't this work fine (i.e. remove all individuals that are now missing from capture data). It shouldn't be necessary to apply this only to conflicted individuals because all other cases will have matching individual and capture records.

```suggestion
      dplyr::filter(!.data$IndvID %in% output_capture$IndvID)
```

R/subset_datarqst.R
```diff
@@ -37,66 +45,141 @@
...
-  save_path <- dirname(file)
+  #If removing conflicted species: remove IndvIDs that do not appear in individual data from capture data
+  #(this ensures that individuals identified as different species do not appear in capture data at all, even if
+  #they were identified as one of the species of interest at one or several captures.)
+  if(!include_conflicting){
+    output_capture <- output_capture %>%
+      dplyr::filter(.data$IndvID %in% output_individual$IndvID)
+  }
+  # NOTE: This comes with an assumption that - aside from individuals with conflicting species - all
+  #       individuals that appear in capture data also appear in individual data (they should)
```
LiamDBailey, line 134: One thing that's still missing here is if an individual is conflicted but only one of the two conflicted species was included in the filters. In this case we will only return part of the capture records for that individual (i.e. only those where it was classified as the species on interest). This doesn't really make sense to me. If a user says they want to include conflicted individuals I would interpret that to mean they want all records for an individual where species is uncertain (regardless of what species it was classified as at the time).

LiamDBailey: I think you could overcome this by first returning a vector of individual IDs from the `Individual data` table and then use individual ID to filter capture ID, rather than using PopID and Species. That way all records for a conflicted species will be returned.

tests/testthat/test_subset_datarqst.R
```diff
@@ -0,0 +1,57 @@
+context("Test that subset_datarqst() functions as expected...")
+
+test_PopIDs <- c('CHO', 'BAN', 'VEL', 'SSQ', 'HOC', 'PIL', 'EDM')
+test_data <- run_pipelines(PopID = test_PopIDs, save = FALSE)
```
LiamDBailey, line 4: We have a `pipeline_output` object that is created when we build the tests. You should use this for the tests rather than calling `run_pipelines` again, just so the tests don't take too long.

LiamDBailey (changes requested):
A few changes that I think might be useful. Most importantly, I think we need to change how we filter 'Capture data' when we're dealing with individuals of conflicted species. We want to return all the capture records, even those where the observed sex may not be one of the focal species.

R/subset_datarqst.R
ChloeRN: @StefanVriend as far as I understand, no it should not. The RDS file that is written by run_pipelines() and read in here is a workspace containing an object called "standard_data". So there should not be any need to assign that here.

R/subset_datarqst.R
ChloeRN: @LiamDBailey in general, I would agree with you that yes, we always want to save to path when we are running the subsetting function for its purpose, i.e. subsetting the data for a data request. 

However, when I was writing the tests for the subsetting function, I realized that when we run the subsetting function in order to test its functionality, then we do NOT want it to save any data into the path. If we did that, then every time we run tests, the "testing subset" will be created anew (overwriting any other subset we may have done on this day). 

However, there may be better ways to fix that perhaps. Open to suggestions!

R/extract_paradox_db.R
```diff
@@ -10,11 +10,25 @@
...
+    #NOTE: This path is specific to Liam as a user. Should verify that this actually does anything
+    # and if so, it should be generalized.
+  }
 
   reticulate::py_run_file(system.file("extdata", "paradox_extract.py", package = "pipelines", mustWork = TRUE))
 
-  output_file <- py$extract_paradox(path = path, file_name = file_name)
+  #Write full pathname
+  # When running on Mac OS, the path in extract_paradox_db may not contain "~"
+  path_full <- paste0(dirname(path), "/", basename(path))
```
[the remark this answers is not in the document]
ChloeRN: Seems to work fine on Mac too. Thanks for the suggestion, didn't know this function existed!

R/subset_datarqst.R
```diff
@@ -37,66 +45,141 @@
 subset_datarqst <- function(file = file.choose(),
                             PopID = unique(pop_names$code),
                             Species = unique(Species_codes$Code),
-                            filter = NULL){
+                            filter = NULL,
+                            include_conflicting = FALSE,
+                            output_type = "R",
+                            return_R = FALSE,
```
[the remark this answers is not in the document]
ChloeRN: I agree, it's best to keep it as similar as possible. Checked back in run_pipelines, and it looks like there the R object is always returned. So, I have removed the "return_R" argument here, and made it so that subset_datarqst always returns the R object too.

R/subset_datarqst.R
ChloeRN: Following a suggestion also from Stefan, I have now removed the `output_R` argument here.

R/subset_datarqst.R
ChloeRN: @LiamDBailey I have now updated the function description to clarify this to the user.

R/subset_datarqst.R
LiamDBailey: In the file `teardown-xyz.R` in the \test folder you can remove any files that were made during the test process. If you make the function save files in a different location (i.e. not overwriting any existing files) and then delete them after tests are done this would remove that issue.

R/subset_datarqst.R
ChloeRN: Great suggestion @StefanVriend, your solution returns the identical result but is much easier to read. Have implemented it :-)

tests/testthat/test_subset_datarqst.R
ChloeRN: Thanks for the tip, I was not aware that we'll have the `pipeline_output`available for this. Have adjusted both the function and the tests accordingly.

R/subset_datarqst.R
ChloeRN: @LiamDBailey I have followed your suggestion, and the filtering now works as follows: 
1) Filter Individual data (if `include_conflicted = TRUE` then *all* individuals with `Species = 'CCCCCC'`or `CONFLICTED` will remain in the data)
2) Filter Capture data using the `IndvID`still present in Individual data
3) Identify all conflicted species individuals that were *never* observed as one of the species of interest, and remove them from the data. 

Step 3) is there to make sure that if someone requested data on blue tits and great tits, including conflicting species individuals, their output data will contain all individuals that have been identified at least once as a great tit or a blue tit, but it will not include individuals that have never been identified as either of this species (so an individual once recorded as a pied flycatcher, and once as a collared flycatcher, would not be included in the output data). 

Personally, I think this does make sense. 
Alternatively, we could just return all conflicting species individuals (irrespective of what they were identified as) in all cases. 
What do you think makes most sense?

R/subset_datarqst.R
ChloeRN: Filtering now works differently, so this has been removed.

R/subset_datarqst.R
LiamDBailey: I think that makes sense. From the users perspective when they ask for (for example) great tit and conflicted species I would take that to mean "return all individuals that could _possibly_ be great tit". So an individual that has at one time been recorded as a great tit is relevant (even if it was also recorded as something else), but an individual only ever recorded as pied/collared flycatcher is not.

ChloeRN: Great. I have added code to make sure that will be handled as discussed, and further specified this in the function documentation.

R/subset_datarqst.R
ChloeRN: @LiamDBailey that's good to know regarding `teardown_xyz.R`. However, I do not really think that it can solve the problem here: if we want to save files to a different location during testing, then we still need some kind of `save`variable to specify the alternative save path...
I think I may just struggle a bit with thinking outside the box on this one. Could you check out the branch and implement your preferred solution? 
(I'll check back on it afterwards, so I can learn what works in this situation).

ChloeRN:
@LiamDBailey and @StefanVriend thanks for the constructive feedback, I have now taken into account almost all of it, so requesting re-review for you to check the improvements I made. 

@StefanVriend requested to add some progress bars to the subsetting function to visualize the progress. 
At the moment, I keep getting warning messages that progress bars are now deprecated. So not quite sure if
and how (and where) to include progress bars in the function at this point. 
Opinions?

StefanVriend:
The function `progess::progress_bar` works quite fine! You can find an example in `check_values_brood` in "brood_check.R".

When I run `subset_datarqst(PopID = "HOG")` and then select the location of the .RDS file, I still run into the following error:
`Error in eval(lhs, parent, parent) : object 'standard_data' not found`. Do I do something wrong?

R/subset_datarqst.R
zuzulaz: I was wondering if it will return also the species name which it was in conflict with? For example for conflicted great tit, return in other column (?) that it was also stated as blue tit in certain cases? Would that be of interest of users or would  they get this information in some other place?

ChloeRN: @zuzulaz which "potential" species an individual has been recorded as is not part of Individual data. It would be a bit cumbersome to have in a single or multiple columns, since - in theory - an individual could be observed as not just two but many different species (hopefully not!). 
However, if the user goes into Capture data and looks for the conflicted species individual there, then each capture will be associated with a species. 

For example: 

In Individual data, the species of individual X is denoted as "CCCCCC" .
In Capture data, individual X has two captures, one of which has Species = "PARMAJ" and one has Species = "CYACAE".

ChloeRN:
@StefanVriend I double-checked the issue with reading in the standard data now, and it looks like you were indeed correct. Now changed the code to what you suggested earlier. Let me know if it runs fine now.

R/subset_datarqst.R
zuzulaz: Right! Thank you for reminding me that that information is in the capture data.

StefanVriend:
@ChloeRN Yes :), it worked now. However, it took 55 minutes. Do you know why it might have taken so long?

ChloeRN:
@StefanVriend I had a look into it, and what takes longest is the part of the function that goes through each individual to determine if its Species is conflicting, and if so, whether the observed species are relevant for the output. I have made two changes here: 
1) Added a filter so that the function only checks "relevance" of observed species if the species is recorded as conflicting in individual data (this should drastically reduce the runtime). 
2) Added a progress bar - as you suggested - to this part of the function (although after changing 1), it's so short that it does not even display...)

Additionally, I've added some messages into the function that give the user an overview over the steps. 

Please try again, and let me know if it works better/faster now.

R/subset_datarqst.R
```diff
@@ -30,73 +42,189 @@
...
 
     output_individual <- standard_data$Individual_data %>%
-      dplyr::mutate(Pop_sp = paste(PopID, Species, sep = "_")) %>%
-      dplyr::filter(Pop_sp %in% filter)
+      dplyr::mutate(Pop_sp = paste(.data$PopID, .data$Species, sep = "_")) %>%
+      dplyr::filter(.data$Pop_sp %in% PopSpec_Filter) %>%
+      dplyr::select(-.data$Pop_sp)
+
+    output_capture <- standard_data$Capture_data %>%
+      dplyr::filter(.data$IndvID %in% output_individual$IndvID)
```
StefanVriend, line 91: I've ran the code again on HOG and this time it took 47 minutes. 
I took a closer look at the function and it seemed that this filter (and the filter on l. 112-113) is the time-consuming part. Then I was wondering, it is possible to first apply a PopID filter and then start cycling through the IndvIDs? I applied the filter below and it only took 8 minutes. Could this be a solution to the speed problem or am I missing something?

```suggestion
    output_capture <- standard_data$Capture_data %>%
      dplyr::filter(.data$CapturePopID %in% {{PopID}} & .data$ReleasePopID %in% {{PopID}}) %>%
      dplyr::filter(.data$IndvID %in% output_individual$IndvID)
```

R/subset_datarqst.R
```diff
@@ -30,73 +42,189 @@
...
     output_brood <- standard_data$Brood_data %>%
-      dplyr::filter(PopID %in% {{PopID}} & Species %in% {{Species}})
+      dplyr::filter(.data$PopID %in% {{PopID}} & .data$Species %in% {{Species}})
+
+    output_individual <- standard_data$Individual_data %>%
+      dplyr::filter(.data$PopID %in% {{PopID}} & .data$Species %in% {{Species}})
 
     output_capture <- standard_data$Capture_data %>%
-      dplyr::filter(CapturePopID %in% {{PopID}} & Species %in% {{Species}})
+      dplyr::filter(.data$IndvID %in% output_individual$IndvID)
```
StefanVriend, line 113: See comment on l. 90-91.

R/subset_datarqst.R
```diff
@@ -30,73 +42,189 @@
...
+
+    output_capture <- output_capture %>%
+      filter(!(.data$IndvID %in% Irrelevant_IndvIDs$IndvID))
+  }
+
+  #Combine output into one R object (and return it)
+  output_data <- list(Brood_data = output_brood,
+                      Capture_data = output_capture,
+                      Individual_data = output_individual,
+                      Location_data = output_location)
```
StefanVriend, line 188: When looking at the output of the subsetting function, I wondered whether we should re-assign Row in each table, so that it starts at 1 rather than the lowest row number the subset had in the .RDS of the standard data.

I also wondered whether we should order the tables by a date or year column before returning the subsetted data rather than the unique identifier of that table.

StefanVriend (commented):
Thanks! The code was still slow when tested on HOG (47 minutes), but I might have found a way to reduce the time a lot; see suggestion below.

R/subset_datarqst.R
StefanVriend: Another thought: maybe we should also _group by_ PopID, as IndvIDs might not be unique across populations (especially of different species).

LiamDBailey: This is definitely an issue worth considering. We don't require IndvIDs to be unique across populations so we shouldn't assume this will be the case.

R/run_pipelines.R
```diff
@@ -65,6 +65,31 @@ run_pipelines <- function(path = choose_directory(),
...
+                    toString(PopID[which(PopID%in%PopID_Access)]),
+                    ". To obtain standard format data for these populations, please run on a Windows OS.")
+              )
+    }
+
+    PopID <- PopID[which(!(PopID%in%PopID_Access))]
+
+  }else{
+    stop(paste0('Operating system ', OS, ' not supported'))
+  }
```
LiamDBailey: This will currently fail for any non Mac system! 😨 There needs to be an `if(windows){}` chunk.

R/subset_datarqst.R
```diff
@@ -8,16 +8,28 @@
...
 #'
 #' @param file The location of the file in the standard format (.RDS file). This file is created using `run_pipelines`.
 #' @param PopID Character vector of three letter population codes. Include all populations that are requested.
 #' @param Species Character vector of six letter species codes. Include all species that are requested.
-#' @param filter Character vector of unique population species combinations (in the format PopID_Species).
-#' Include all unique population species combinations requested.
-#'
-#' @return Create 4 .csv files at the location of the standard data. Files will be given a unique number
-#' based on the date when the subset was performed.
+#' @param PopSpec_Filter Character vector of unique population species combinations (in the format PopID_Species).
```
LiamDBailey, line 18: Being pendantic here, but if this one is called 'PopSpec_Filter' then shouldn't the others be 'PopID_Filter' and 'Species_Filter' 🤔. Maybe just 'PopSpec' would be better.

R/subset_datarqst.R
```diff
@@ -30,73 +42,189 @@
...
+  }
+
+  #Combine output into one R object (and return it)
+  output_data <- list(Brood_data = output_brood,
+                      Capture_data = output_capture,
+                      Individual_data = output_individual,
+                      Location_data = output_location)
+
+  message(crayon::green("Data subset returned as R object."))
+  return(output_data)
```
LiamDBailey, line 191: ⚠️ This return is hit every time you run the code, so all the lines below this, including the `output_type` and `save` argument, are ignored.

R/subset_datarqst.R
```diff
@@ -30,73 +42,189 @@
...
+
+
+    #Save output as RDS or .csv files
+    if(output_type == "R"){
+
+      saveRDS(output_data, file = paste0(save_path, "/Output_data_", as.character(Sys.Date()), ".RDS"))
+
+      message(crayon::cyan(crayon::bold("Data subset saved as R data file (.RDS).")))
+
+    }else{
```
LiamDBailey, line 213: Besides the above issue, this will also save as csv with *any* other value of `output_type`, even e.g. if the user writes `output_type = 'r'`. I think we want to explicitly check that `output_type = 'csv'` before we save any .csv files.

```suggestion
}else if(output_type == "csv"){
```

R/subset_datarqst.R
```diff
@@ -30,73 +42,189 @@
...
+                            save = TRUE,
+                            test = FALSE){
+
+  if(!test){
+    message(crayon::bold(crayon::green("Select standard data file (RDS format).")))
+    standard_data <- readRDS(file = file)
+  }else{
+    message(crayon::green("Using 'pipeline_output' as standard data."))
+    standard_data <- pipeline_output
+  }
```
LiamDBailey, line 64: This should be a fine workaround for now, but might cause some issues in `devtools::check()`. I will deal with this later though.

R/run_pipelines.R
```diff
@@ -65,6 +65,31 @@ run_pipelines <- function(path = choose_directory(),
...
+    if(length(PopID[which(PopID%in%PopID_Access)] > 0)){
+      warning(paste0('Pipelines not run for the following populations due to OS incompatibility: ',
+                    toString(PopID[which(PopID%in%PopID_Access)]),
+                    ". To obtain standard format data for these populations, please run on a Windows OS.")
+              )
+    }
+
+    PopID <- PopID[which(!(PopID%in%PopID_Access))]
+
+  }else if(!grepl(pattern = 'mac|windows', x = OS)){
```
LiamDBailey, line 85: I've changed this because before it was failing for *all* non-Mac systems (including windows 😨)

LiamDBailey (changes requested):
Looks much better. I think your workaround to use in tests should be fine, although I'll need to see how this works with `devtools::check()` (I'll worry about that). *However*, there is one major bugs that still need to be fixed before merging. Currently, the return on l.191 in `subset_datarqst.R` will end the function so that the arguments `save` and `output_type` are being completely ignored (i.e. it will only ever return an object inside R).

R/subset_datarqst.R
ChloeRN: Changed as suggested.

R/subset_datarqst.R
ChloeRN: Moved the return to the end of the function, which should hopefully take care of that.

R/subset_datarqst.R
ChloeRN: Taken care of this.

R/subset_datarqst.R
ChloeRN: @StefanVriend as also discussed in person, yes if this helps with speed then it makes sense to include it. Have added it now (with some modifications that take into account filtering by `PopSpec`vs. filtering by `PopID`and `Species`.

R/subset_datarqst.R
ChloeRN: @StefanVriend regarding the row number, this we need to discuss and I will send a mail to the dev team. 

As for the ordering of tables, I agree with you that it makes sense to order the tables in a more intuitive/natural way. 
I am now ordering tables as follows: 
Brood: PopID > BreedingSeason > Species > BroodID
Capture: CapturePopID > IndvID > BreedingSeason > CaptureDate > CaptureTime
Individual: PopID > Species > IndvID
Location: PopID > LocationType > StartSeason > EndSeason > LocationID

This ordering is what I personally think people may be most used to. Can easily be changed though if someone has strong opinions. 

(If we are going to re-assign row numbers, we should do that after ordering the tables I think).

R/subset_datarqst.R
ChloeRN: @StefanVriend and @LiamDBailey, good point about IndvID potentially not being unique across populations. 

I figured there were 2 ways to address this: 1) make sure IndvID is unique by ensuring this at the pipeline level or 2) take into account that IndvID is linked to PopID when subsetting. 
The problem with 1) is that this will not be able to correctly deal with migrants and transplanted individuals. 

Therefore, I went with 2). Stefan suggested grouping by `PopID` while subsetting by `IndvID`, but I could not think of a good way to do that. Instead, I introduce a temporary variable `PopID_IndvID` and use this during the subsetting process. This may not be the most elegant solution, but it works nicely.

StefanVriend:
Looks good. 9 min and almost 14 seconds for HOG is a big improvement compared to the 45-minute running time.

If we all agree that Rows should be reset to 1, that piece of code should indeed be added after each table is arranged as you suggested above.

LiamDBailey (changes requested):
We still need to make a few more changes to work with conflicted individuals properly. Code will currently only return the capture records for a conflicted species that match the focal species (while we want all capture records). More detail with an example in my review comments. Pushed some test code that should be passed for things to be working.

I've also pushed some changes that return a useful error if we accidentally add an unknown PopID or Species.

Unfortunately I couldn't run this on all the pops because my NIOO account has been disabled and needs to be re-activated 🙄

R/subset_datarqst.R
```diff
@@ -108,16 +112,20 @@ subset_datarqst <- function(file = file.choose(),
       dplyr::filter(.data$PopID %in% {{PopID}} & .data$Species %in% {{Species}})
 
     output_individual <- standard_data$Individual_data %>%
+      dplyr::mutate(PopID_IndvID = paste(.data$PopID, .data$IndvID, sep = "_")) %>%
       dplyr::filter(.data$PopID %in% {{PopID}} & .data$Species %in% {{Species}})
 
     output_capture <- standard_data$Capture_data %>%
-      dplyr::filter(.data$IndvID %in% output_individual$IndvID)
+      dplyr::filter(.data$CapturePopID %in% {{PopID}}  & .data$Species %in% {{Species}}) %>%
```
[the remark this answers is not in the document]
ChloeRN: Great catch! I've now accepted your suggestion and did a test-run on the entire data, and it seems to work smoothly now.

R/subset_datarqst.R
```diff
@@ -64,35 +64,39 @@ subset_datarqst <- function(file = file.choose(),
...
+                    PopID_IndvID = paste(.data$PopID, .data$IndvID, sep = "_")) %>%
+      dplyr::filter(.data$PopID_Species %in% {{PopSpec}}) %>%
+      dplyr::select(-.data$PopID_Species)
 
     output_capture <- standard_data$Capture_data %>%
-      dplyr::filter(.data$CapturePopID %in% {{PopID}} & .data$ReleasePopID %in% {{PopID}}) %>%
-      dplyr::filter(.data$IndvID %in% output_individual$IndvID)
+      dplyr::mutate(PopID_Species = paste(.data$CapturePopID, .data$Species, sep = "_"),
+                    PopID_IndvID = paste(.data$CapturePopID, .data$IndvID, sep = "_")) %>%
+      dplyr::filter(.data$PopID_Species %in% {{PopSpec}}) %>%
```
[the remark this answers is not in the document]
ChloeRN: Made a change equivalent to what you suggested above, and this now seems to work as well.

ChloeRN:
Made final adjustments following Liam's suggestions and everything seems to work smoothly now. Ran tests and RCMD checks, all were passed. Ready to merge now.

LiamDBailey:
@ChloeRN I'm having some problem getting the tests and checks to work on my side. Could you show me the output of the tests and checks to compare?

PR #139, opened 2020-12-04
Pew pipeline

zuzulaz:
I uploaded what I have done so far. There are two files,  one where I basically only copied the general structure of the function from other pipelines. The other file is temporal one, where I work on the functions.
So far is quite messy and with a lot of notes, I will clean that as they get checked and clarified.


R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#' #### ---------------------------------------------------------------------------------------~
+#' #### Questions/Doubts DATA OWNERS
+#' # What is "Neophobia Transponder"?
+#' # Code "Experiment"?
+#' # OriginalTarsusMethod
+#' # Check experimentID
+#'
+#' #### Questions/Doubts TEAM
+#' # What to do with strange ID, like "MetalRing&Tag_M2019NestK55" (34 individuals)?
+#' # What do do with individuals without ID?
```
LiamDBailey, line 41: - If brood parents are unknown this can just be left blank.
- If an unknown individual is captured and the capture still includes useful information (e.g. mass, tarsus) then I would create a unique ID for unknown individuals that is clearly distinct from regular IDs (e.g. UNK_0001). This way the valid data from this individual capture could still be used (e.g. determine avg mass of individuals in year X) and we would still fill the IndvID column. 
- If an unknown individual is captured and no useful information is collected, then the record can simply be removed (but make sure to specify that these individuals were removed in the help doc). This seems to be the case for the 4 records of chick-ringing with no IndvID.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#' # Where does the chick-rearing data go? To Individual data?
```
LiamDBailey, line 42: What do you mean by 'chick-rearing'? I couldn't see this in the data. If you're referring the to 'ChickRinging' records, these should be treated as captures (i.e. they are the first capture of an individual during their lifetime).

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'
+#' #### ---------------------------------------------------------------------------------------~
+#'
+#'
+#'
+#'
+#' library(pipelines)
```
LiamDBailey, line 49: It might be easier to build this as a function from the start so you can add in the help documentation and record decisions you make as you work (e.g. exclude unknown individuals). That will also remove the need to comment out everything before you save.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'
+#'
+#' db_dir <- choose_directory() ## temp copy in C:\Users\ZuzanaZ\Dropbox\POSTDOC\POSTDOC_NIOO_NETHERLANDS_2020\SPI-birds_project\data_pipeline_temp_copy\PEW_PEWrdsbosWest_Belgium
```
LiamDBailey, line 52: We need to keep the arguments identical between functions, so this should be `db`

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'
+#' #### --------------------------------------~
+#' #### INDIVIDUAL DATA                    #### done, need to check
+#' #### --------------------------------------~
+#'
+#' #### Still to solve:
+#' # There are some individuals (for example 13617043), where the rows of data related
+#' # to the same individual seem complementary, but are in two rows with
+#' # insufficient information when treating the rows separately.
+#' # How to deal with that?
```
LiamDBailey, line 137: - Seems to me that there are some records that are not captures (i.e. they don't have a date) but just include brood information. So, I would explicitly remove the records with no date from Capture_data but still use them in Brood_data
- In other cases where there are multiple captures in a year (with dates) these should be treated as legitimate multiple captures.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'
+#'
+#' create_individual_PEW <- function(data){
+#'
+#'   Individual_data <-
+#'     data %>%
+#'     #### Remove rows without bird ID
+#'     filter(!is.na(IndvID)) %>%
+#'     #### CHECK: Remove rows without date
+#'     # filter(!is.na(Date)) %>%
```
LiamDBailey, line 161: See comment above, I think this is correct of Capture_data (and corresponding Individual_data) 👍 😄

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'     #### Format and create new data columns
+#'     dplyr::select(IndvID, Species, PopID, Age, Method, BreedingSeason, Date, Sex) %>%
+#'     dplyr::rename(RingSeason = BreedingSeason) %>%
+#'     dplyr::mutate(RingAge = dplyr::case_when(Age == "0" & Method == "ChickRinging" ~ "chick",
+#'                                              Age == "1" ~ "adult",
+#'                                              Age == ">=2" ~ "adult"),
+#'                   BroodIDLaid    = NA_character_,
```
LiamDBailey, line 168: We have data on the BroodID for individuals caught with method 'ChickRinging', so in those cases this should not be NA.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'                                                               return("C")
+#'                                                             }
+#'                                                           }),
+#'                   Sex_genetic = NA_character_) %>%
+#'     dplyr::ungroup() %>%
+#'     #### Final arrangement
+#'     dplyr::select(IndvID, Species, PopID, BroodIDLaid, BroodIDFledged,
+#'                   RingSeason, RingAge, Sex_calculated, Sex_genetic) %>%
+#'     #### Remove duplicated rows (CHECK)
+#'     dplyr::distinct()
```
LiamDBailey, line 186: The Individual_data table should include a *single record for each individual* with information that should remain static throughout their life (e.g. sex). The PEW primary data is structured more like our Capture_data, with *multiple records for each individual* with information that changes for each capture (e.g. mass, tarsus). To achieve the Individual_data output we'll need to group_by()/summarise() for each individual, otherwise we'll end up with multiple records for each individual (e.g. look at 13617015 as an example). I'd recommend run `create_capture_PEW` first and then use this as the basis for the `create_individual_PEW`. 😄

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'   #### Create new variables
+#'   mutate(#### ASSUMPTION: There is no captureID info, so far I use this (CHECK)
+#'          CaptureID = paste(IndvID, "CaptureNumber", sep = "_"),
+#'          #### ASSUMPTION: Some times are strange, to simplify, round to whole hour
+#'          CaptureTime = ifelse(is.na(ObservationTimeH),
+#'                               NA,
+#'                               paste(sprintf("%02d", round(ObservationTimeH, 0)),
+#'                                     "00", sep = ":")),
+#'          Sex_observed = Sex,
+#'          LocationID = NA_character_,
```
LiamDBailey, line 221: We know the NestBox where each individual is captured, which can be used to fill in the LocationID

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'
+#'
+#'
+#' # }
+#'
+#'
+#'
+#' #### VARIOUS ASSUMPTIONS:
+#' # For Age: - there are some idividuals with NA age, but with partner and the DateRingingChick
+#' #             >> we assume they are adults (example IndvID 12706844)
```
LiamDBailey, line 262: Exactly how I would approach it. Generally, if there is no age specified we should assume they're adult because this is much more likely than being an unrecorded chick!

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#' # For CaptureDate: - there are some individuals with NA in Date, but in some there is DateRingingChick,
+#' #                    can we use that and assume that as Date?
```
LiamDBailey, line 264: Would double check with data owner but I assume this is the date that chicks from that clutch were ringed, not the date that the parents were ringed. If so, it is not the same as 'Date'.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'
+#' create_location_PEW <- function(data){
+#'
+#'   Location_data <-
+#'     data %>%
+#'     dplyr::select(BreedingSeason, Date, NestboxID, PopID) %>%
+#'     group_by(NestboxID) %>%
+#'     arrange(BreedingSeason, Date) %>%
+#'     dplyr::mutate(StartSeason = first(BreedingSeason),
+#'                   EndSeason = last(BreedingSeason)) %>%
```
LiamDBailey, line 399: Check with data owner. If nest boxes are not taken down we usually keep this as NA.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'     select(-BreedingSeason, -Date) %>%
+#'     dplyr::distinct() %>%
```
LiamDBailey, line 401: If you use summarise rather than mutate you won't need to remove duplicates.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,419 @@
...
+#'     dplyr::ungroup() %>%
+#'     dplyr::mutate(LocationID = NestboxID, # CHECK
+#'                   LocationType = "NB",
+#'                   Latitude  = 51.266667,
+#'                   Longitude = 4.466667,
```
LiamDBailey, line 406: Ask data owner if they have this info (they often do but just don't share it 😓).

LiamDBailey (changes requested):
Be careful when creating Individual_data, it should contain a single row per individual. Left some comments to point out some potential issues and tips.

zuzulaz:
Regarding the Individual data, the multiple records were caused by the RingSeason variable, where I kept all years, not just the first one, I will correct that. 
One thing, in the protocol, the RingSeason is defined as "The breeding season (year) when an individual was first _ringed_.", what I find a bit confusing, as it seems to me it refers more generally to "was first captured". In particular in individuals, that were ringed previously and we do not have the detailed information in the data available about when it was.

R/format_PEW_temp.R
zuzulaz: Yes, sorry, I ment "ChickRinging".

R/format_PEW_temp.R
zuzulaz: ok

R/format_PEW_temp.R
zuzulaz: Yes, I changed that.

R/format_PEW_temp.R
LiamDBailey: Exactly, you can just create one. Generally something like 'Year_Nestbox_LayDate' will work to create a unique ID as it accounts for renesting in the same nestbox during the same year.

LiamDBailey:
I would assume in most (all?) populations that there are not individuals that were ringed before their first capture, because that would mean they were either a) ringed by somebody else at the same site or b) immigrated from another site where birds were being ringed. I guess this is possible at PEW because PEE is close, but I would assume that for every bird the first capture date is also the first ringing date. Maybe check with the data owner when you meet with them. If he did not ring all the birds himself then we need to find out who did and where we can get that data (possibly they were first ringed in PEE)

R/format_PEW_temp.R
zuzulaz: Ok, I changed it to NA, added to questions for data owner.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+    #### Rename variables
+    dplyr::rename(LocationID = NestboxID) %>%
+    dplyr::select(-DateRingingChicks) %>%
+    #### Create new variables
+    dplyr::mutate(#### ASSUMPTION: There is no captureID info, so I made this up
+                  CaptureID = paste(IndvID, LocationID,
+                                    paste0(lubridate::year(CaptureDate),
+                                           sprintf("%02d", lubridate::month(CaptureDate)),
+                                           sprintf("%02d", lubridate::day(CaptureDate))),
+                                    sep = "_"),
```
LiamDBailey, line 386: From the standard protocol v1.1.0: 'CaptureID is a concatenation of `IndvID` and capture number.'. Possible code to meet this requirement below.

```suggestion
    dplyr::group_by(IndvID) %>% 
    dplyr::mutate(CaptureID = paste(IndvID, row_number())) %>%
    dplyr::ungroup() %>%
    dplyr::mutate(
```

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+      #### ASSUMPTION: Some times are strange, to simplify, round to whole hour
+      #### ? do they correspond only to the experimental observations? keep/remove?
+      CaptureTime = ifelse(is.na(ObservationTimeH),
+                           NA,
```
LiamDBailey, line 390: Better to specify. If we try to run the pipeline on a subset of the data where ObservationTimeH is always missing this will return a logical NA rather than a character (which doesn't fit our standard protocol).

```suggestion
                           NA_character_,
```

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+      CaptureAlive = ifelse(Method != "Found dead", TRUE, FALSE),
+      #### ASSUMPTION: the same as CaptureAlive, as there is no additional data
+      ReleaseAlive = CaptureAlive,
+      CapturePopID = PopID,
+      CapturePlot = NA_character_,
+      ReleasePopID = ifelse(ReleaseAlive == TRUE, CapturePopID, NA_character_),
+      ReleasePlot = ifelse(ReleaseAlive == TRUE, CapturePlot, NA_character_),
+      #### ASSUMPTION: use NA, as there is no information
+      OriginalTarsusMethod = NA_character_,
+      WingLength = NA,
```
LiamDBailey, line 403:
```suggestion
      WingLength = NA_real_,
```

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+      ExperimentID = dplyr::case_when(Experiment == "BSM - Griffioen et al. 2019 PeerJ" ~
+                                        "COHORT; PARENTAGE",
+                                      Experiment == "2h Temp D4" ~ "SURVIVAL",
+                                      Experiment == "2h Temp D4 + Handicaping Male: Griffioen et al. 2019 Front Ecol&Evol" ~
+                                        "SURVIVAL; PARENTAGE"),
```
LiamDBailey, line 408: Brood ExperimentID and Capture ExperimentID don't necessarily need to be the same. For example, a cross foster experiment will affect the parentage of a brood and the chicks, but not of the parents.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                               Age == "1" ~ 5,
+                               Age == ">=2" ~ 7,
+                               #### Account for controls in December
+                               Age == "1" & Method == "Night Control Winter" &
+                                 lubridate::month(CaptureDate) == 12 ~ 6,
+                               #### Account for controls in January & February
+                               Age == ">=2" & Method == "Night Control Winter" &
+                                 lubridate::month(CaptureDate) %in% c(1, 2) ~ 8,
+                               is.na(Age) ~ 5),
+      Age_observed = as.integer(Age_observed))
```
LiamDBailey, line 420: Can explicitly make a number an integer with `L`

```suggestion
      Age_observed = case_when(Age == "0" & Method == "ChickRinging" ~ 1L,
                               Age == "1" ~ 5L,
                               Age == ">=2" ~ 7L,
                               #### Account for controls in December
                               Age == "1" & Method == "Night Control Winter" &
                                 lubridate::month(CaptureDate) == 12 ~ 6L,
                               #### Account for controls in January & February
                               Age == ">=2" & Method == "Night Control Winter" &
                                 lubridate::month(CaptureDate) %in% c(1, 2) ~ 8L,
                               is.na(Age) ~ 5L))
```

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+      OriginalTarsusMethod = NA_character_,
+      WingLength = NA,
+      ExperimentID = dplyr::case_when(Experiment == "BSM - Griffioen et al. 2019 PeerJ" ~
+                                        "COHORT; PARENTAGE",
+                                      Experiment == "2h Temp D4" ~ "SURVIVAL",
+                                      Experiment == "2h Temp D4 + Handicaping Male: Griffioen et al. 2019 Front Ecol&Evol" ~
+                                        "SURVIVAL; PARENTAGE"),
+      #### NOT SURE ABOUT THIS
+      Age_observed = case_when(Age == "0" & Method == "ChickRinging" ~ 1,
+                               Age == "1" ~ 5,
```
LiamDBailey, line 411: Always worth checking how age translates into our age codes (e.g. does "1" mean that we explicitly know the birth year of the individual?). Always an important question to ask the data owner 🕵️

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                               Age == ">=2" ~ 7,
```
LiamDBailey, line 412: '>=2' means we don't know the age with certainty, so this should be 6 ("After 2nd year: full-grown bird hatched before last calendar year; year of hatching otherwise unknown.") rather than 7 ("3rd year: a bird hatched two calendar years before, and now in its third calendar year.") We have no way to tell if the bird is in its third calendar year or later.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                               #### Account for controls in December
+                               Age == "1" & Method == "Night Control Winter" &
+                                 lubridate::month(CaptureDate) == 12 ~ 6,
+                               #### Account for controls in January & February
+                               Age == ">=2" & Method == "Night Control Winter" &
+                                 lubridate::month(CaptureDate) %in% c(1, 2) ~ 8,
+                               is.na(Age) ~ 5),
```
LiamDBailey, line 419: If age is unknown we have to use an age category with uncertainty, so this should be 4 ("After first-year: full-grown bird hatched before this calendar year; year of hatching otherwise unknown") rather than 5 ("2nd year: a bird hatched last calendar year and now in its second calendar year"). We have no way to tell if a bird with Age = NA is in it's second calendar year or older.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+      #### NOT SURE ABOUT THIS
+      Age_observed = case_when(Age == "0" & Method == "ChickRinging" ~ 1,
+                               Age == "1" ~ 5,
+                               Age == ">=2" ~ 7,
+                               #### Account for controls in December
+                               Age == "1" & Method == "Night Control Winter" &
+                                 lubridate::month(CaptureDate) == 12 ~ 6,
+                               #### Account for controls in January & February
+                               Age == ">=2" & Method == "Night Control Winter" &
+                                 lubridate::month(CaptureDate) %in% c(1, 2) ~ 8,
```
LiamDBailey, line 418: What about Age 1 individuals caught in January/February or Age >=2 individuals caught in December?

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+
+  # AGE_CALCULATED SEEMS NOT CORRECT ?
+  # (maybe also because the Age_observed was not correctly assigned?)
+  #
+  # Liam:
+  #   Seems to me that there are some records that are not captures (i.e. they don't have a date)
+  #   but just include brood information. So, I would explicitly remove the records with no date
+  #   from Capture_data but still use them in Brood_data.
+  #   In other cases where there are multiple captures in a year (with dates)
+  #   these should be treated as legitimate multiple captures.
```
LiamDBailey, line 143: Generally, yes I agree. CaptureDate is required in the Capture_data table. However, in this case it does seem like there's some useful capture information in the records without dates. e.g. in the first row of the data, we can assume that 11462816 was observed at some time during 2015 because it was classified as the female for nestbox 1. This type of observation still tells us something about things like the age of the individual at subsequent captures (we already know that it was an adult in 2015). So I would:

- Where an individual is observed as a brood parent and is not captured otherwise, include that record as a capture (give it a random date in 2015, maybe 01-01-2015, and just be explicit that this is a made-up date).
- Where an individual has a separate capture record with a date, ignore the records without dates.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                     RingSeason = first(BreedingSeason),
+                     RingAge = first(RingAge),
+                     BroodIDLaid = first(BroodIDLaid),
+                     BroodIDFledged = first(BroodIDFledged),
+                     Sex_calculated = first(Sex_calculated),
+                     Sex_genetic = first(Sex_genetic)) %>%
+    dplyr::ungroup() %>%
+    #### Final arrangement
+    dplyr::select(IndvID, Species, PopID, BroodIDLaid, BroodIDFledged,
+                  RingSeason, RingAge, Sex_calculated, Sex_genetic)
```
LiamDBailey, line 506: `Sex_calculated` is just intended to return a single value per individual, so `summarise` would be more appropriate. Your current code doesn't work for some individuals. For example, 13617071 was caught as a chick in nest 2016_91, but it is not given this as BroodIDLaid because the CaptureDate of this record is NA (but see my comment above. The below code should be more robust.

 ```suggestion
Individual_data <- data %>% 
    group_by(IndvID) %>% 
    dplyr::summarise(Sex_calculated = purrr::map_chr(.x = list(unique(na.omit(Sex_observed))), .f = ~{
        if(length(..1) == 0){
            return(NA_character_)
        } else if(length(..1) == 1){
            return(..1)
        } else {
            return("C")
        }
    }),
    Sex_genetic = NA_character_,
    Species = first(Species),
    PopID = "PEW",
    #IndvID = first(IndvID), #Not needed because it's the grouping factor
    RingSeason = min(BreedingSeason),
    RingAge = ifelse(min(Age_observed) == 1, "chick", "adult"),
    BroodIDLaid = ifelse(RingAge == "chick",
                         paste(BreedingSeason, LocationID, sep = "_"),
                         NA_character_),
    BroodIDFledged = BroodIDLaid) %>%
    dplyr::ungroup() %>% 
    dplyr::select(IndvID, Species, PopID, BroodIDLaid, BroodIDFledged,
                  RingSeason, RingAge, Sex_calculated, Sex_genetic)
```

File: R/format_PEW_temp.R [hunk 26]
```diff
@@ -0,0 +1,541 @@
...
+    #### Final arrangement
+    dplyr::select(CaptureID, IndvID, Species, Sex_observed, BreedingSeason,
+                  CaptureDate, CaptureTime, ObserverID, LocationID,
+                  CaptureAlive, ReleaseAlive, CapturePopID, CapturePlot,
+                  ReleasePopID, ReleasePlot, Mass, Tarsus, OriginalTarsusMethod,
+                  WingLength, Age_observed, Age_calculated, ChickAge, ExperimentID)
+
+    # #### Remove records with no date
+    # #### CHECK: doing this, almost 500 records of chicks will be removed,
+    # #### as there is no CaptureDate !!!
```
LiamDBailey, line 451: Many of the chick records do have a capture date, they just need to be calculated from the data. e.g. 13617071 has a capture record where the date is 'D14', which I assume is 14 days after hatching (need to ask the data owner to be sure), which would give it a capture date of 2016-05-28 (i.e. the hatch date of the brood 2016-05-14 + 14)

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+create_location_PEW <- function(data) {
+
+  Location_data <-
+    data %>%
+    dplyr::select(BreedingSeason, Date, NestboxID, PopID) %>%
+    group_by(NestboxID) %>%
+    arrange(BreedingSeason, Date) %>%
+    dplyr::summarise(StartSeason = first(BreedingSeason),
+                     #### CHECK WITH DATA OWNER
+                     #### EndSeason = last(BreedingSeason))
```
LiamDBailey, line 521: BreedingSeason is an integer, so you can just use min/max rather than arrange/first. arrange/first has the risk of failing if there is an NA Date, because these won't arrange properly.

```suggestion
    dplyr::summarise(StartSeason = min(BreedingSeason),
                     #### CHECK WITH DATA OWNER
                     #### EndSeason = max(BreedingSeason))
```

LiamDBailey (changes requested):
@zuzulaz great work 👍 Most of my comments are just things that could make the code cleaner/more robust. A few bigger things that still need to be fixed:

- **Deal with NA capture date**: This dataset is a bit unusual. Usually we would ignore records with no date, but there is some usable information here that is missing somewhere else. Where an individual is only ever observed in a record with no date, we should create a capture record with a made up date so that we record that it was observed in that breeding season. For chicks where date is DXXX, we need to calculate capture date as HatchDate + XXX.
- **Double check `Age_observed`**: Remember that when age is unknown (e.g. >=2 or NA) we need to use an age category which includes uncertainty. See pg. 14 in the EURING data https://euring.org/files/documents/E2000PLUSExchangeCodeV117.pdf
- **Create tests**: There are a few cases that would be good to create tests for that are currently not working in the code (e.g. individual 13617071 should have BroodIDLaid 2016_91_PEW not NA, because it was caught as a chick).

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                                               "text", "text", "text", "text", "text",
+                                               "text", "numeric", "date", "date",
+                                               "text", "date", "numeric", "numeric",
+                                               "text", "numeric", "numeric", "numeric",
+                                               "numeric", "numeric", "numeric",
+                                               "numeric", "numeric", "numeric")) %>%
+    janitor::clean_names(case = "upper_camel") %>%
+    janitor::remove_empty(which = "rows") %>%
+    #### Solve NA values (some are explicitly stated as character "NA")
+    dplyr::mutate(across(where(is.character), .fns = ~replace(., . ==  "NA" , NA))) %>%
```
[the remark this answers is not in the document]
zuzulaz: Thanks, I was not aware of this argument.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                  HatchDateD0 = as.Date(HatchDateD0),
+                  NumberOfRingedChicks = as.integer(NumberOfRingedChicks),
+                  DateRingingChicks = as.Date(DateRingingChicks),
+                  NoOfChicksD3 = as.integer(NoOfChicksD3),
+                  BroodMassD3 = as.numeric(BroodMassD3),
+                  ObservationTimeH = as.numeric(ObservationTimeH),
+                  Species = "CYACAE",
+                  PopID = "PEW",
+                  NestboxID = tolower(Nest),
+                  BroodID = ifelse(Method %in% c("Catch adults nestbox", "ChickRinging"),
```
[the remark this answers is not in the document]
zuzulaz: I added it. At the end it saves one case from being ignored, and adds one conflict in the Brood data (due to differences in the clutch size in "Catch adults nestbox" and "Catch incubation" for the same BroodID) . I am not sure whether this is very case-specific or will also appear in the future datasets for next years.

"Catch incubation" method BroodID:
2017_107_PEW >> the only record for this brood
2017_109_PEW >> two records for the same brood, one for "Catch incubation" and other for "Catch adults nestbox",with differnt ClutchSize and HatchDate. I would keep only the one for the "Catch adults nestbox", as it contains more information.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                  PopID,
+                  IndvID,
+                  Sex,
+                  Age,
+                  everything()) %>%
+    #### Remove observations without ID
+    dplyr::filter(!is.na(IndvID)) %>%
+    #### Details
+    dplyr::mutate(PartnerId = ifelse(PartnerId == "MetalRing_FemK89_2019",
+                                     "MetalRing_F2019_K89", PartnerId))
```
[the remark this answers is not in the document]
zuzulaz: This might get tricky, as there are several forms of coding for unringed individuals. I first though to write some generalizable code to unify them, but after I only found the one specific individual written in two different forms, I transformed that one in particular.

The most common ways used are:
Unknow_SexNest_Year (Unknown_Male75_2017)
Unknow_SexYearNest (seems to be a version used only in the year 2019: Unknown_M2019NestK97)
u/Unringed_SexNest_Year (unringed_Nest103_2017)
UniParental (or semiparental or Semi-parental or SemiUniParental)_SexNest_Year (UniParental_MaleK99_2016)
MetalRing or NoMetalRing_SexYearNest (MetalRing_M2019Nest103, NoMetalRing_M2019NestK110)

If you think it would be good to unify them, do you have in mind any specfic format? Maybe we could remove the first word (unringed, unknow,...) and keep only the sex, nest and year in unified format, for example Sex_NestID_Year?

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+
+create_brood_PEW <- function(data) {
+
+  parents_brood_data <-
+    data %>%
+    #### Exclude non-breeding data, exclude chicks
+    #### Use only CatchAdultsNestbox, where all relevant information is.
+    #### However, for one nest in the clutch size is only in the CatchIncubation Method !!
+    #### Ignore for now
+    filter(Method == "Catch adults nestbox") %>%
```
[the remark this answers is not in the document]
zuzulaz: See my comment above related to 'Catch incubation' and BroodID. In general there in very little relevant data.

R/format_PEW_temp.R
```diff
@@ -0,0 +1,541 @@
...
+                  HatchDate_max = NA_character_,
+                  BroodSize_observed = NA_integer_,
+                  BroodSize_min = NA_integer_,
+                  BroodSize_max = NA_integer_,
+                  FledgeDate_observed = NA_character_,
+                  FledgeDate_min = NA_character_,
+                  FledgeDate_max = NA_character_,
+                  #### ASSUMPTION: There is no specific information about the
+                  # number of chicks fledged, use the number of chicks ringed
+                  # !? in few cases this number is higher than clutch size
```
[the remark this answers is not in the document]
zuzulaz: Ok, I added that to the questions for data owner. 
Right, I would expect the cases where ringed > clutch size to be marked by Stefan's check, I noted it here to be sure not to miss that.

R/format_PEW_temp.R
zuzulaz: Ok, I added the arrange by capture date

   ```
 dplyr::group_by(IndvID) %>% 
    dplyr::arrange(Year, CaptureDate) %>% 
    dplyr::mutate(CaptureID = paste(IndvID, row_number(), sep = "_")) %>%
    dplyr::ungroup() %>%
```

File: R/format_PEW_temp.R (hunk 16 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:390]: Right!

File: R/format_PEW_temp.R (hunk 17 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:403]: Ok.

File: R/format_PEW_temp.R (hunk 18 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:408]: I am not quite sure what you mean. Is it that there should/could be different ExperimentID in the Capture_data and in the Brood_data?

File: R/format_PEW_temp.R (hunk 19 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:420]: Ok.

File: R/format_PEW_temp.R (hunk 20 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:411]: Already on the list of questions 👍

File: R/format_PEW_temp.R (hunk 21 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:412]: Ok, I have always found this way of noting the age confusing :)

File: R/format_PEW_temp.R (hunk 22 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:419]: Ok.

File: R/format_PEW_temp.R (hunk 23 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:418]: I would say that individuals caught in December would have the same Age_observed as individuals caught in May within the same calendar year. Therefore it doesn't need a specific condition, right?

Regarding the Age 1 individuals caught in January/February, I think there is an error in the code (I had also the options for the December, which I later removed, but it seems that I deleted the right condition).

```
#### Account for controls in January & February
                               Age == "1" & Method == "Night Control Winter" &
                                 lubridate::month(CaptureDate) %in% c(1, 2) ~ 6L,
```

File: R/format_PEW_temp.R (hunk 27 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:521]: Ok.

File: R/format_PEW_temp.R (hunk 26 above)
Comment (zuzulaz) [reply R/format_PEW_temp.R:451]: Oh, I totally missed that "D14" information! I assume is that after the transformation of the column to date, all non-date characters were changed to NAs... 🤦‍♀️
Sure, now I can do the calculation from the hatching date. Thanks!

File: R/format_PEW_temp.R (hunk 18 above)
Comment (LiamDBailey) [reply R/format_PEW_temp.R:408]: Exactly. Some experiments will affect the brood and parents differently. For example, supplemental feeding of adults might improve adult survival but only change brood phenology.

**Review by LiamDBailey — CHANGES_REQUESTED** [review]
Some small changes hopefully streamline code/make it more robust

File: R/format_PEW.R [hunk 33]
```diff
@@ -24,77 +68,466 @@ format_PEW <- function(db = choose_directory(),
...
+                                               "text", "text", "numeric", "text",
+                                               "text", "text", "text", "text", "text",
+                                               "text", "text", "text", "text", "text",
+                                               "text", "text", "text", "text", "text",
+                                               "text", "text", "text", "text",
+                                               "text", "text", "numeric", "numeric",
+                                               "text", "numeric", "numeric", "numeric",
+                                               "numeric", "numeric", "numeric",
+                                               "numeric", "numeric", "numeric"),
+                                 na = "NA") %>%
```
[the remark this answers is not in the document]
zuzulaz: Right, changed.

R/format_PEW.R
```diff
@@ -24,77 +68,466 @@ format_PEW <- function(db = choose_directory(),
...
+                                                               date_system = "modern"),
+                  Tarsus = as.numeric(Tarsus),
+                  Mass = as.numeric(Mass),
+                  ClutchSize = as.integer(stringr::str_replace(string = ClutchSize,
+                                                               pattern = "\\?",
+                                                               replace = "")),
+                  NumberOfRingedChicks = case_when(DateRingingChicks %in% c("verlaten incubatie", "all dead d1") ~ 0L,
+                                                   HatchDateD0 %in% c("abandoned", "bumblebee nest", "dead embryos") ~ 0L,
+                                                   NumberOfRingedChicks == "all dead" & Nest %in% c("66", "70") ~ 0L,
+                                                   NumberOfRingedChicks == "all dead" & Nest == "79" ~ 10L,
```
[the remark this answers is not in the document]
zuzulaz: That is case specific, I added the year, what I think would other possibilities in future datasets.
NumberOfRingedChicks == "all dead" & Nest == "79"  & "BreedingSeason" == 2015 ~ 10L

R/format_PEW.R
```diff
@@ -0,0 +1,533 @@
...
+                                                   TRUE ~ as.integer(NumberOfRingedChicks)),
+                  HatchDateD0 = janitor::excel_numeric_to_date(as.numeric(HatchDateD0),
+                                                               date_system = "modern"),
+                  DateRingingChicks = janitor::excel_numeric_to_date(as.numeric(DateRingingChicks),
+                                                                     date_system = "modern"),
+                  NoOfChicksD3 = as.integer(NoOfChicksD3),
+                  BroodMassD3 = as.numeric(BroodMassD3),
+                  NewRing = toupper(NewRing),
+                  Species = "CYACAE",
```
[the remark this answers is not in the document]
zuzulaz: Changed.

R/format_PEW.R
```diff
@@ -0,0 +1,533 @@
...
+    dplyr::summarise(StartSeason = min(BreedingSeason, na.rm = TRUE),
+                     EndSeason = NA_integer_,
+                     LocationID = unique(NestboxID),
+                     PopID = unique(PopID)) %>%
+    dplyr::ungroup() %>%
+    dplyr::mutate(LocationType = "NB",
+                  HabitatType = "deciduous",
+                  #### Only general coordinates for the location
+                  Latitude  = 51.266667,
+                  Longitude = 4.466667) %>%
```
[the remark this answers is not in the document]
zuzulaz: ok

R/format_PEW.R
```diff
@@ -0,0 +1,533 @@
...
+    dplyr::mutate(HatchDate_observed = if_else(is.na(HatchDate_observed) & is.na(CaptureDate) & Method == "ChickRinging",
+                                               as.Date(paste0(BreedingSeason, "-06-01")),
+                                               HatchDate_observed),
+                  CaptureDate = case_when(is.na(CaptureDate) & D14Chicks == "yes" ~ HatchDate_observed + 14,
+                                          #### Create fake capture dates
+                                          is.na(CaptureDate) & Method == "Catch adults nestbox" ~
+                                            as.Date(paste0(BreedingSeason, "-06-01")),
+                                          is.na(CaptureDate) & Method == "Catch incubation" ~
+                                            as.Date(paste0(BreedingSeason, "-04-01")),
+                                          TRUE ~ CaptureDate)) %>%
```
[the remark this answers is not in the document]
zuzulaz: Ok, I set both to
` is.na(CaptureDate) ~ as.Date(paste0(BreedingSeason, "-04-01"))`

R/format_PEW.R
```diff
@@ -0,0 +1,533 @@
...
+                     NewRing, CaptureDate, NeophobiaTransponder,
+                     Tarsus, Mass, ObserverID, Experiment, D14Chicks)) %>%
+    #### Remove duplicated rows (as we get one row for males and females for the same brood)
+    dplyr::distinct() %>%
+    #### Remove rows with no information about the brood
+    dplyr::filter(!(is.na(ClutchSize) & is.na(DateEgg1) & is.na(HatchDateD0) &
+                      is.na(NumberOfRingedChicks) & is.na(DateRingingChicks) &
+                      is.na(NoOfChicksD3) & is.na(BroodMassD3))) %>%
+    #### Remove one specific case causing double register for the same BroodID
+    dplyr::filter(!(BroodID == "2017_109" & Method == "Catch incubation")) %>%
```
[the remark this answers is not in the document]
zuzulaz: I have asked data owner.

R/format_PEW.R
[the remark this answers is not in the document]
zuzulaz: I have checked again about this case, related to your other comment below for the Brood data. The thing is that in this nest 79, 10 chicks were ringed (they appear in the individual data actually), although after few days they were found dead. So I will simplify the code and change all cases in the variable NumberOfRingedChicks  "all dead" to 0, as actually I only use this variable for assigning the NumberFledged.

PR #141, opened 2020-12-14
Archive update

LiamDBailey:
I found a few extra changes needed for the archive function when I was doing the initial archiving. I've pushed the final code and also **adding an archiving folder to all populations with primary data on the N drive.** Feel free to look through the code, but it would be even more helpful to check on the N-drive to see if you can spot any mistakes in the archiving output.


ChloeRN:
Hi @LiamDBailey! I will look through the N-drive when I go to my office on Wednesday (can only check via my Desktop computer there since I need to "tunnel" through a virtual PC from there).

LiamDBailey:
@ChloeRN probably better because it's already quite slow to check at the best of times. The PFN should at least be functioning, so you should be able to continue with the data updates.

LiamDBailey:
@ChloeRN Added a bit about the archive process into the README too.

ChloeRN:
@LiamDBailey I had a chance to look at the N-drive now. I checked for the presence of archives for all populations, and double-checked contents and ArchiveMetaData for some. For the largest part, it looks like the archiving works fine. Here are a few specific observations of things that may need to be fixed/improved/thought about:

1. At the moment, GOTL and NIOO do not have an archive. These are the only populations with 4-letter owner codes... so that could be where the problem lies?

2. Some populations had already received data updates (e.g. HAR, KEV, SSQ). The presence of update data in a folder did not mess with the archiving, but we just have to keep in mind that these "older" updates will need a tidy-up at some point. 

3. So far, only PrimaryData gets archived. However, when people add more years to their raw data, they will presumably also update MetaData? So perhaps MetaData should also be archived together with PrimaryData?

Aside from that, it looks like it's working! (Did not look through code yet...)

zuzulaz:
Hi, I haven't gone through the code, however, I checked the files on Ndrive. I have a question regarding the names of folders with dates  (for example, ".../archive/2020-10-01"). I would expect them to be the date when the data were archived (or is it not the case?), but for different populations, the names of this folder differ (but majority seems to be some 3-4 different  dates).

LiamDBailey:
@zuzulaz this will be the case in the future, but here I'm just creating the initial archive for each population so I used the date at which we got the data originally (and when we should have archived it...if we had the archive code back then 😅)

@ChloeRN thanks for checking! This is a problem with the loop I wrote to apply the archiving not the archive code itself. The archive code should be used one population at a time. NIOO data should be added now and GOTL no longer exists (it's just incorporated into the GOT folder)

Issue #140, opened 2020-12-11
Non-standard formats in pipelines_output


ChloeRN:
The pipelines_output for all datasets contains formatting inconsistencies which probably arise from one or a few pipelines. 
These include:
- Columns with "named" elements
- Remaining grouping within columns
- Columns that are not part of the standard format

In some select cases, this also causes issues with tests. Should be sorted out.

PR #152, opened 2021-01-21
PFN masterpipeline

ChloeRN:
format_PFN() is now updated to the "master pipeline", which formats data for East Dartmoor (EDM), Dinas (DIN), Loch Katrine (KAT), Forest of Dean (NAG), Okehampton (OKE), and Teign (TEI).

Besides formatting data from the other populations, the main new feature of the pipeline is that it consolidated information contained in the two sources of raw data ("Nest data" and "IPMR data"). 

RCMD checks currently on-going, but will confirm once passed.


ChloeRN:
RCMD checks passed on my branch.

(Had to disable the test for column presence/order again as it does not work on the output of `run_pipelines()` at the moment. Issue #140 will have to be fixed first, and then we need a better way to implement this test. I.e. one that can take into account different versions of the standard format.)

R/format_PFN.R
```diff
@@ -190,60 +371,50 @@ format_PFN <- function(db = choose_directory(),
...
+                                             .data$Species %in% c("GRETI", "GRETI ") ~ species_codes[species_codes$SpeciesID == 14640, ]$Species,
                                              .data$Species == "REDST" ~ species_codes[species_codes$SpeciesID == 11220, ]$Species,
                                              .data$Species == "MARTI" ~ species_codes[species_codes$SpeciesID == 14400, ]$Species,
                                              .data$Species == "NUTHA" ~ species_codes[species_codes$SpeciesID == 14790, ]$Species,
                                              .data$Species == "COATI" ~ species_codes[species_codes$SpeciesID == 14610, ]$Species,
-                                             .data$Species == "WREN" ~ NA_character_, # Missing, 1 observation only
-                                             .data$Species == "TREEC" ~ NA_character_), # Missing, 1 observation only
+                                             .data$Species == "WREN" ~ NA_character_, # Not currently included, 1 observation only
+                                             .data$Species == "TREEC" ~ NA_character_), # Not currently, 1 observation only
```
LiamDBailey, line 417: If you think these could be valid later might be worth adding to `species_codes`

R/format_PFN.R
```diff
@@ -336,216 +517,407 @@ create_brood_EDM <- function(Primary_data, ReRingTable){
...
+        is.na(.data$CapturePlot_IPMR) ~ .data$CapturePlot_Nest,
+        .data$CapturePlot_Nest == .data$CapturePlot_IPMR ~ .data$CapturePlot_Nest,
+        .data$CapturePlot_Nest != .data$CapturePlot_IPMR ~ .data$CapturePlot_Nest
+      ),
+      ReleasePlot = dplyr::case_when(
+        is.na(.data$ReleasePlot_Nest) ~ .data$ReleasePlot_IPMR,
+        is.na(.data$ReleasePlot_IPMR) ~ .data$ReleasePlot_Nest,
+        .data$ReleasePlot_Nest == .data$ReleasePlot_IPMR ~ .data$ReleasePlot_Nest,
+        .data$ReleasePlot_Nest != .data$ReleasePlot_IPMR ~ .data$ReleasePlot_Nest
+      ),
```
LiamDBailey, line 837: This rule could be written out as a single function so it doesn't need to be retyped each time. From a quick check it seems you're using the same decision rule every time (i.e. use Nest data where possible)

R/calc_birdNBuse.R
```diff
@@ -16,25 +17,29 @@
...
 #' library(dplyr)
-#' Brood_data <- tibble::tibble(BroodID = c(1:10), LocationID = LETTERS[c(1,3,1,1,2,3,1,2,1,2)],
-#'                              BreedingSeason = sort(sample(2012:2016, 10, replace = TRUE)))
-#' calc_birdNBuse(Brood_data)
+#' Brood_data <- tibble::tibble(BroodID = rep(c(1:5), 2),
+#'                              PopID = rep(c('XXX', 'YYY'), each = 5),
+#'                              LocationID = LETTERS[c(1,2,1,1,2,2,1,2,1,2)],
+#'                              BreedingSeason = c(2001, 2001, 2002, 2003, 2003,
+#'                              1962, 1963, 1963, 1965, 1965))
+#'calc_birdNBuse(Brood_data)
```
LiamDBailey, line 31: Could you include something like this as a small test for `calc_birdNBuse` (and also include a test for the ReRinging code too).

LiamDBailey (changes requested):
Some small changes, biggest thing to check is the reringing code. I'm running the tests/checks on my side atm and will let you know how it goes.

R/calc_birdNBuse.R
```diff
@@ -16,25 +17,29 @@
...
 
   output <- Brood_data %>%
-    dplyr::group_by(LocationID) %>%
-    dplyr::summarise(StartSeason  = as.integer(min(BreedingSeason)),
-                  EndSeason  = as.integer(max(BreedingSeason))) %>%
+    dplyr::group_by(.data$PopID, .data$LocationID) %>%
+    dplyr::summarise(StartSeason  = as.integer(min(.data$BreedingSeason)),
+                  EndSeason  = as.integer(max(.data$BreedingSeason)),
+                  .groups = 'keep') %>%
     dplyr::ungroup()
```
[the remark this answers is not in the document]
ChloeRN: Great @LiamDBailey, thanks for this. You may have noticed my comment about something not working quite right yet with the `make_IndIdentifier()` function. If I recall correctly, issues arose when there were complimentary matches. For example, a bird with 2 captures, ringed at capture 1 with ring number "X1" and re-ringed at capture 2 to new ring number "X2", but both ring numbers reported like: 
CaptureNo           RING            RING2
1                           X1                X2
2                          X2                X1

I had already changed the function to prevent an "infinite loop" earlier, but could not figure out how to prevent it from including both complimentary pairs (X1-X2 and X2-X1) in the output.

R/format_PFN.R
```diff
@@ -121,52 +241,113 @@ format_PFN <- function(db = choose_directory(),
     dplyr::group_by(.data$Identifier) %>%
     dplyr::mutate(ReRingID = first(.data$RingNr)) %>%
     dplyr::ungroup() %>%
-    dplyr::distinct(.data$RingNr, .data$ReRingID)
+    dplyr::distinct(.data$RingNr, .data$ReRingID) %>%
+    dplyr::distinct(.data$RingNr, .keep_all = T)
+    # NOTE: At the moment, I am using this last line to remove duplicate pairings
+    #       arising from two-way matches in RingCombos. Eventually, this is something
+    #       that the make_IndIdentifier() function should fix.
```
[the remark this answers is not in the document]
ChloeRN: See my comment above. I think it's related to that.

R/format_PFN.R
ChloeRN: Most of the time, but not always. Priority is given to IPMR for example for `CaptureAlive` and `ReleaseAlive`, and for some columns, information is consolidated differently. 

(Not quite sure how I would write it as a single function. Without actually writing a function...)

R/format_PFN.R
```diff
@@ -67,33 +73,147 @@ format_PFN <- function(db = choose_directory(),
...
+  #--------------------------------#
+  # ASSIGNING INDIVIDUAL IDENTITIY #
+  #--------------------------------#
+  # This is necessary because some birds have carried several rings during their lifetime (re-ringed)
 
   # Extract ring number (combinations) for all birds
-  RingCombos <- CMR_data[,c('RING', 'RING2', 'DATE')] %>%
+  RingCombos <- IPMR_data[,c('RING', 'RING2', 'DATE')] %>%
     dplyr::mutate(DATE = as.Date(.data$DATE, format = "%d/%m/%Y")) %>%
     dplyr::distinct(.data$RING, .data$RING2, .keep_all = T)
```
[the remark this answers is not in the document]
ChloeRN: Good point. I technically know that but it's good to get a reminder!

R/format_PFN.R
ChloeRN: I don't think so. These are the "richest" datasets PiedFlyNet administers (as I understand) and there's really not many records here...

R/format_PFN.R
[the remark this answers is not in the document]
LiamDBailey: Hmm ok, do you have a reproducible example where it wasn't working? If the `make_IndIdentifier` isn't working is there a reason we can't just use the code directly from RingCombos I suggested above?

R/format_PFN.R
LiamDBailey: So just something like this would work, and you could add some variation to work for other columns

```
compare_IPMR_Nest <- function(Nestcol, IPMRcol){
         
             dplyr::case_when(is.na({{Nestcol}}) ~ {{IPMRcol}},
                                                     is.na({{IPMRcol}}) ~ {{Nestcol}},
                                                     {{Nestcol}} == {{IPMRcol}} ~ {{Nestcol}},
                                                     {{Nestcol}} != {{IPMRcol}} ~ {{Nestcol}})
}

#Now this would work. Saves 5 lines of code
Capture_data %>%
mutate(CapturePopID = compare_IPMR_Nest(.data$CapturePopID_Nest, .data$CapturePopID_IPMR)
```

R/calc_birdNBuse.R
ChloeRN: I've made a small test for `calc_birdNBuse`. Test for ReRinging code we should make after fixing the function.

PR #156, opened 2021-02-05
MTA pipeline

zuzulaz:


R/format_MTA.R
```diff
@@ -0,0 +1,371 @@
...
+
+  #### Primary data
+  message("Importing primary data...")
+
+  mta_data <- readxl::read_excel(path =  paste0(db, "/MTA_PrimaryData.xlsx"),
+                                 na = c("NA", "nA")) %>%
+    janitor::clean_names(case = "upper_camel") %>%
+    janitor::remove_empty(which = "rows") %>%
+    #### Convert to corresponding format and rename
+    dplyr::mutate(PopID = "MTA",
```
LiamDBailey, line 67: See comment above about PopID

R/format_MTA.R
```diff
@@ -0,0 +1,371 @@
...
+#' interest or necessity of the data user. Currently, only general coordinates
+#' for the location are provided.
+#'
+#' @inheritParams pipeline_params
+#' @return Generates either 4 .csv files or 4 data frames in the standard format.
+#' @export
+
+
+#### --------------------------------------------------------------------------~
+#### QUESTIONS DATA OWNERS
```
LiamDBailey, line 25: Antica's comment from the SPI-Birds spreadsheet:

"their data is in hungarian, but my understanding is that he has sent a translated version.... check before working on this"

Should check with data owner when you contact them.

R/format_MTA.R
```diff
@@ -0,0 +1,371 @@
...
+    tidyr::pivot_longer(cols = c(.data$FemaleID, .data$MaleID),
+                        names_to = "Sex_observed",
+                        values_to = "IndvID") %>%
+    dplyr::mutate(Sex_observed = substr(.data$Sex_observed, start = 1, stop = 1)) %>%
+    #### Remove records where the partner is not known
+    dplyr::filter(!is.na(.data$IndvID)) %>%
+    #### Remove unringed individuals
+    dplyr::filter(.data$IndvID != "unringed") %>%
+    #### Create capture date
+    dplyr::mutate(CaptureDate = as.Date(paste0(BreedingSeason, "-06-01"))) %>%
```
LiamDBailey, line 247: Check with data owners

R/format_MTA.R
```diff
@@ -0,0 +1,371 @@
...
+#' @param Brood_data Data frame. Brood_data from MTA-PE Evolutionary Ecology Group, Hungary.
+#' @return A data frame.
+#'
+
+
+create_capture_MTA <- function(data) {
+
+  Capture_data <-
+    data %>%
+    dplyr::select(.data$Species, .data$PopID, .data$BreedingSeason,
```
LiamDBailey, line 234: We still want to calculate species from the `species_code` table in case we want to change the codes in the future.

R/format_MTA.R
```diff
@@ -0,0 +1,371 @@
...
+#' This section provides details on data management choices that are unique to
+#' this data. For a general description of the standard format please see see
+#'\href{https://github.com/SPI-Birds/documentation/blob/master/standard_protocol/SPI_Birds_Protocol_v1.0.0.pdf}{here}.
+#'
+#' \strong{xxx}:
+#'
+#' \strong{Latitude, Longitude} The exact coordinates of the nestboxes are not available.
+#' Data owner provided an map of the location, which can be georeferenced in the case of
+#' interest or necessity of the data user. Currently, only general coordinates
+#' for the location are provided.
```
LiamDBailey, line 17: Is this still true?

LiamDBailey (changes requested):
A few important things:

- Looks like the MTA data includes 5 different populations (sites), so the PopID should vary between records.
-  Make sure we check that this is the raw data/they can recreate this data. Just like with KIL it looks like they've converted directly to our standard format.

R/format_MTA.R
```diff
@@ -0,0 +1,371 @@
+#' Construct standard format data from MTA-PE Evolutionary Ecology Group, University of Pannonia, Hungary (MTA)
+#'
+#' A pipeline to produce the standard format for bird study population
+#' at the MTA-PE Evolutionary Ecology Group, Hungary, administered by
+#' Gabor Seres & András Liker.
+#' The data include 5 sites: Veszprém, Balatonfüred, Szentgál, Vilma-puszta and Gulya-domb.
```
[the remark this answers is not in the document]
zuzulaz: Ok. Should I therefore remove the MTA population?

R/format_MTA.R
```diff
@@ -0,0 +1,371 @@
...
+#'
+#' This section provides details on data management choices that are unique to
+#' this data. For a general description of the standard format please see see
+#'\href{https://github.com/SPI-Birds/documentation/blob/master/standard_protocol/SPI_Birds_Protocol_v1.0.0.pdf}{here}.
```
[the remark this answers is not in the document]
zuzulaz: 👍

R/format_MTA.R
zuzulaz: Yes, I have seen that Antica's comment. The only word in Hungarian I found is the "gyurutlen" which based on dictionary refers to unringed bird. Appears only several times, majority of unringed individuals are labeled "ringless" or "unringed" in English.

R/format_MTA.R
zuzulaz: In this data there is column called Species with "PARMAJ" value. Should I change it to extract the species from the `species_code `instead?

R/format_MTA.R
zuzulaz: Nope, my mistake. The true is that we do not have coordinates for nestboxes, neither we do have any map. I will ask data owners about that.

R/format_MTA.R
LiamDBailey: Yes, that would be the best approach. In the future we might change the codes (especially if we start getting many different species from overseas), so we want to have it linked to the EURING code.

R/format_MTA.R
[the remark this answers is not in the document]
LiamDBailey: Exactly. MTA can be kept as an owner, but is no longer a valid PopID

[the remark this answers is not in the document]
zuzulaz: Done.

LiamDBailey:
@zuzulaz Made a few minor changes, but otherwise seems fine. I will leave this PR open for when we get the new data, but that will be done by @cwtyson.

zuzulaz:
Ok, great, thanks. It means that I can mark this pipeline as finished for me? :)

Issue #159, opened 2021-02-18
Adjust save path of quality check report


LiamDBailey:
- When running for a single dataset, it would be good to save in the data folder AND ARCHIVE!
- When running for a data request, it would be good to create a new folder with subset data and save quality check there.

Issue #158, opened 2021-02-18
Add summary of each check in the quality check report


LiamDBailey:
Should allow the data owner/user to understand why a record has been flagged. Particularly important for comparison to reference values, we can give the reference value to the reader.

Issue #157, opened 2021-02-18
Check age of subsequent captures


LiamDBailey:
There are cases where an individual may be listed as adult and then chick when ordered chronologically, obviously wrong (e.g. 14154750 in PEW pipeline). To deal with this we could:

- Individual is chick on the same year or later than it was an adult (ERROR)
- Age !>= previous age chronologically (WARNING)

PR #161, opened 2021-02-19
Quality check testrun

StefanVriend:
This PR involves:

- [x] - Add new check C5: checking that the age of subsequent captures is in the correct order (#157)
- [x] - Add check descriptions to the quality check report (#158)
- [x] - Add reference values to quality check report (#158)
- [ ] - Change save path of quality check function (#159)


StefanVriend:
All tests in `devtools::test()` passed.
@LiamDBailey Shall we look at the save path (#159) before reviewing this PR & merging into the master?

LiamDBailey:
@StefanVriend let's leave that for later. Shouldn't affect output given to the data owner so we can do that later, maybe when we talk about infrastructure. I'll check the tests are also running on my side and then we can merge.

LiamDBailey:
Spotted a few small issues that had to be corrected:
- Included documentation for the new B13/14 tests (and removed reference to check_sex_parents) 📜 
- Be careful when copying text from outside of R because you can introduce non-ASCII characters (which R can't deal with). The most common problem is with quotation marks where these slanted quotes are non-ASCII: ‘’ but these straight ones are fine: ''. There were a few cases of this in the new quality check documentation that I spotted 🕵️.

Running `devtools::check()` as well as `devtools::test()` will help to spot these type of things.

StefanVriend:
Thanks, @LiamDBailey!

In my case, `devtools::test()` did not highlight them, but `devtools::check()` did indeed (which I forgot to run before I mailed you).

PR #166, opened 2021-04-09
Adding format_GLA

cwtyson:
Rough draft pipeline for Glasgow data.


tests/testthat/test_GLA_pipeline.R
```diff
@@ -0,0 +1,124 @@
...
+  expect_equal(subset(GLA_data, IndvID == "AFE3005")$Sex_observed[2], "M") # Male
+  expect_equal(subset(GLA_data, IndvID == "AFE3005")$LocationID[2], "545") # Nest 65
+  expect_equal(min(subset(GLA_data, IndvID == "AFE3005")$CaptureDate, na.rm = TRUE), as.Date("2020-04-30"))
+
+  #Test 2: Female caught as adult and at three different nests up to and including 2020
+  #Test the female has the correct number of capture records (3) AXB1206
+  expect_equal(nrow(subset(GLA_data, IndvID == "AXB1206")), 3) # Two records
+  #Test the second capture of the male
+  expect_equal(subset(GLA_data, IndvID == "AXB1206")$Sex_observed[1], "F") # Female
+  expect_equal(subset(GLA_data, IndvID == "AXB1206")$LocationID[2], "543") # Nest 543
```
LiamDBailey, line 88: Second capture was still in the Fledging nest (513). See line 1594 in the excel spreadsheet.

tests/testthat/test_GLA_pipeline.R
```diff
@@ -0,0 +1,124 @@
...
+  expect_equal(min(subset(GLA_data, IndvID == "AXB1206")$CaptureDate, na.rm = TRUE), as.Date("2018-05-08"))
```
LiamDBailey, line 89: First capture should be '2018-05-25'

tests/testthat/test_GLA_pipeline.R
```diff
@@ -0,0 +1,124 @@
...
+
+})
+
+test_that("Location_data returns an expected outcome...", {
+
+  #Take a subset of only GLA data
+  GLA_data <- dplyr::filter(pipeline_output$Individual_data, PopID %in% "CAS", "GAR", "KEL", "SAL", "SCE") %>%
+    arrange(.data$IndvID, .data$BreedingSeason)
```
LiamDBailey, line 97: Be careful when copy/pasting code. IndvID and BreedingSeason don't exist in Location_data

```suggestion
  GLA_data <- dplyr::filter(pipeline_output$Location_data, PopID %in% c("CAS", "GAR", "KEL", "SAL", "SCE"))
```

tests/testthat/test_GLA_pipeline.R
```diff
@@ -0,0 +1,124 @@
...
+
+  #Test 1: Nestbox
+  #LocationType is as expected
+  expect_equal(subset(GLA_data, LocationID == "728")$LocationType, "NB")
+  #Habitat is as expected
+  expect_equal(subset(GLA_data, LocationID == "728")$HabitatType, "urban")
+  #Start season
+  expect_equal(subset(GLA_data, LocationID == "728")$StartSeason, 2015)
```
LiamDBailey, line 105: 2015 appears correct in the raw data, but code is not returning this value. Appears to be a bug :bug:

tests/testthat/test_GLA_pipeline.R
```diff
@@ -0,0 +1,124 @@
...
+  #End season
+  expect_equal(subset(GLA_data, LocationID == "728")$EndSeason, NA_integer_)
+
+  #Test 2: Nestbox
+  #LocationType is as expected
+  expect_equal(subset(GLA_data, LocationID == "123")$LocationType, "NB")
+  #Habitat is as expected
+  expect_equal(subset(GLA_data, LocationID == "123")$HabitatType, "deciduous")
+  #Start season
+  expect_equal(subset(GLA_data, LocationID == "123")$StartSeason, 2014)
```
LiamDBailey, line 117: As above :bug:

R/data_table_templates.R
```diff
@@ -0,0 +1,105 @@
...
+                                PopID = NA_character_, ## Missing data not allowed
+                                BroodIDLaid = NA_character_,
+                                BroodIDFledged = NA_character_,
+                                RingSeason = NA_integer_,
+                                RingAge = NA_character_,
+                                Sex_calculated = NA_character_,
+                                Sex_genetic = NA_character_)
+
+## Save
+save(individual_data_template, file = "./data/Individual_data_template.Rda")
```
LiamDBailey, line 44: It's annoying, but R expects .rda (i.e. lowercase) rather than .Rda in the data folder, even though they are functionally the same. 

```suggestion
save(individual_data_template, file = "./data/Individual_data_template.rda")
```

R/data_table_templates.R
```diff
@@ -0,0 +1,105 @@
...
+                              NumberEggs = NA_integer_,
+                              AvgChickMass = NA_real_,
+                              NumberChicksMass = NA_integer_,
+                              AvgTarsus = NA_real_,
+                              NumberChicksTarsus = NA_integer_,
+                              OriginalTarsusMethod = NA_character_,
+                              ExperimentID = NA_character_)
+
+## Save
+save(brood_data_template, file = "./data/Brood_data_template.Rda")
```
LiamDBailey, line 88:
```suggestion
save(brood_data_template, file = "./data/Brood_data_template.rda")
```

R/data_table_templates.R
```diff
@@ -0,0 +1,105 @@
...
+                                 LocationType = NA_character_, ## Missing data not allowed
+                                 PopID = NA_character_, ## Missing data not allowed
+                                 Latitude = NA_real_,
+                                 Longitude = NA_real_,
+                                 StartSeason = NA_integer_,
+                                 EndSeason = NA_integer_,
+                                 HabitatType = NA_character_)
+
+## Save
+save(location_data_template, file = "./data/Location_data_template.Rda")
```
LiamDBailey, line 105:
```suggestion
save(location_data_template, file = "./data/Location_data_template.rda")
```

LiamDBailey (changes requested):
@cwtyson Thanks for all the work so far 👍, I've included comments and suggestions in the code and you can reply there if you need more clarification. The full function `GLA_format()` is currently not working on my side and the output doesn't seem to pass the tests. This seems to be due to the argument names being used inside the function, which need fixing. I've gone through and reviewed the code as best I could, but in the future make sure the code is stable on your side before the review process.

R/format_GLA.R
```diff
@@ -0,0 +1,638 @@
...
+    dplyr::filter(is.na(.data$LocationID) == F) %>%
+
+    ## Determine whether a chick or adult - done for each observation since there is no unique identifier for individuals
+    dplyr::mutate(RingAge = ifelse(.data$Age_observed == 1, "chick", "adult")) %>%
+
+    ## Keeping chicks
+    dplyr::filter(RingAge == "chick") %>%
+
+    ## Summarize brood information for each nest
+    dplyr::group_by(.data$BreedingSeason, .data$PopID, .data$LocationID) %>%
```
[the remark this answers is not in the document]
cwtyson: I agree though I'll need to think about how to do this since it doesn't seem straightforward to join lay date to the ringing data. Something like a rolling join with capture date and lay date could work, but might be messy.

R/format_GLA.R
```diff
@@ -0,0 +1,638 @@
...
+    dplyr::summarise(Species = names(which.max(table(.data$Species, useNA = "always"))),
+                     FemaleID = names(which.max(table(.data$MotherRing, useNA = "always"))),
+                     MaleID = names(which.max(table(.data$FatherRing, useNA = "always"))),
```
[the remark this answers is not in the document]
cwtyson: There are not (double-checked just summarizing by unique values), but I thought in case band numbers were written down incorrectly, then taking the most common name wouldn't break and would be more likely to return the right value. 
Perhaps this is a case though where it is better to get an error so that we know there is an issue?

R/format_GLA.R
```diff
@@ -0,0 +1,638 @@
...
+                                                    "COHORT; PARENTAGE"),
+                  ClutchType_observed = dplyr::case_when(.data$ReplacementClutch == 0 ~ "first",
+                                                         is.na(.data$ReplacementClutch) ~ "first",
+                                                         .data$ReplacementClutch > 0 ~ "second"))
+
+
+  ## Join brood data from ringing records to brood data from nest records
+  ## TODO: Add experiment information
+  Brood_data <- nest_data_brood_sum %>%
+    dplyr::left_join(rr_data_brood_sum, by = c("BreedingSeason", "Species", "PopID", "LocationID")) %>%
```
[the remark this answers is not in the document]
cwtyson: The ring record data only has CaptureDate which probably could be used in some way to join to laying date in the nest data, but it doesn't seem straightforward

R/format_GLA.R
```diff
@@ -0,0 +1,638 @@
...
+                  FemaleID = .data$FemaleID_j) %>%
+
+    ## Keep only necessary columns
+    dplyr::select(dplyr::contains(names(brood_data_template))) %>%
+
+    ## Add missing columns
+    dplyr::bind_cols(brood_data_template[,!(names(brood_data_template) %in% names(.))]) %>%
+
+    ## Reorder columns
+    dplyr::select(names(brood_data_template)) %>%
```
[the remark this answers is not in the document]
cwtyson: These sound like good alternatives. I'm happy to do it whichever way. I do like the idea of using the templates to check the column types.

R/format_GLA.R
```diff
@@ -0,0 +1,638 @@
...
+  Capture_data_rr <- rr_data %>%
+    dplyr::filter(!(.data$IndvID %in% c("too small", "too_small", "no_rings_COVID","unknown"))) %>% # Keep only records of banded individuals
+
+    ## Arrange
+    dplyr::arrange(.data$BreedingSeason, .data$PopID, .data$IndvID, .data$CaptureDate) %>%
+
+    ## Add additional data
+    dplyr::mutate(CapturePopID = .data$PopID, ## Set CapturePopID based on PopID
+                  CaptureAlive = dplyr::case_when(.data$Retrap == "X" ~ FALSE,
+                                                  .data$Retrap %in% c("N", "R", "C", "U", NA) ~ TRUE), ## Set CaptureAlive to FALSE if Retrap is X, otherwise TRUE
```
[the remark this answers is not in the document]
cwtyson: Will do.

R/format_GLA.R
```diff
@@ -0,0 +1,638 @@
...
+    dplyr::select(dplyr::contains(names(capture_data_template))) %>%
+
+    ## Add missing columns
+    dplyr::bind_cols(capture_data_template[,!(names(capture_data_template) %in% names(.))]) %>%
+
+    ## Reorder columns
+    dplyr::select(names(capture_data_template))
+
+
+  ## Create capture data from primary data on brood. This will mostly be used to get information about experiments and perhaps some resightings.
```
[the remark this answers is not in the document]
cwtyson: Yes, there are individuals that listed as parents of a brood, but are not in the capture data for that. Also, there are unique records for individuals by breeding season and capture pop ID (for instance, same individual seen in two populations in the same year). This is another thing worth talking to the data owner about,.

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+                                                                     TRUE ~ lubridate::ymd(paste(unlist(stringr::str_split(as.character(janitor::excel_numeric_to_date(as.numeric(.data$LayingComplete))), pattern = "-"))[c(1,3,2)], collapse = "-"), quiet = TRUE))),
+
+                  ObservedHatch = suppressWarnings(dplyr::case_when(grepl("/", .data$ObservedHatch) ~ lubridate::dmy(.data$ObservedHatch, quiet = TRUE),
+                                                                    TRUE ~ lubridate::ymd(paste(unlist(stringr::str_split(as.character(janitor::excel_numeric_to_date(as.numeric(.data$ObservedHatch))), pattern = "-"))[c(1,3,2)], collapse = "-"), quiet = TRUE))),
+
+                  Hatchlings = as.integer(.data$Hatchlings),
+                  Fledglings = as.integer(.data$Fledglings),
+                  MaleRing = as.character(.data$MaleRing),
+                  FemaleRing = as.character(.data$FemaleRing),
+                  ClutchSize = as.integer(.data$ClutchSize)) %>%
```
LiamDBailey, line 97: Does this mean they've accidentally stored some dates in US format (M/D/Y) and some in non-US format (D/M/Y)? What a nightmare! 😆

This could be a bit streamlined by using `mutate(across())`. Also, after the unusual date issue we can `ungroup()` because it is much slower to do things rowwise. See my suggestion below. 

```suggestion
    dplyr::rowwise() %>%
    
    ## There are two formats for dates.
    ## The first can be handled with Lubridate, the second needs to be rearranged
    dplyr::mutate(dplyr::across(.cols = c(.data$FirstEggDate,
                                   .data$LayingComplete,
                                   .data$ObservedHatch),
                         .fns = ~suppressWarnings(dplyr::case_when(grepl("/", .) ~ lubridate::dmy(., quiet = TRUE),
                                                                   TRUE ~ lubridate::ymd(paste(unlist(stringr::str_split(as.character(janitor::excel_numeric_to_date(as.numeric(.))), pattern = "-"))[c(1,3,2)], collapse = "-"), quiet = TRUE))))) %>%
    dplyr::ungroup() %>%
    dplyr::mutate(
      Hatchlings = as.integer(.data$Hatchlings),
      Fledglings = as.integer(.data$Fledglings),
      MaleRing = as.character(.data$MaleRing),
      FemaleRing = as.character(.data$FemaleRing),
      ClutchSize = as.integer(.data$ClutchSize)) %>% 
```

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+                                           .data$PopID == "sallochy" ~ "SAL",
+                                           .data$PopID == "SCENE" ~ "SCE",
+                                           TRUE ~ as.character(.data$PopID)),
+                  LayDate_observed = .data$FirstEggDate) %>%
+
+    ## Rename
+    dplyr::rename(FemaleID = .data$FemaleRing,
+                  MaleID = .data$MaleRing,
+                  LayDate_min = .data$FirstEggDate,
+                  LayDate_max = .data$LayingComplete,
```
LiamDBailey, line 123: `LayDate` should be the date the first egg was laid, so I don't think we should use `LayingComplete` here. If there is no info on LayDate uncertainty, maybe it's better to just keep these as NA.

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+                  ClutchSize_observed = .data$ClutchSize,
+                  HatchDate_observed = .data$ObservedHatch,
+                  NumberFledged_observed = .data$Fledglings) %>%
+
+    ## Create BroodID based on PopID and row number
+    dplyr::ungroup() %>%
+    dplyr::mutate(BroodID = dplyr::case_when(!is.na(.data$Species) ~ paste(.data$PopID, dplyr::row_number(), sep ="-"))) %>%
```
LiamDBailey, line 130: I would usually include info on BreedingSeason and NestBox, but this is maybe better because you don't need to worry about how to deal with multiple clutches in the same box/year. I would include this in the function description so it's easy for a user to decipher the BroodID values.

R/format_GLA.R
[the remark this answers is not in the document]
LiamDBailey: Do we have a solution for this yet? Could consider using `fuzzyjoin`.

R/format_GLA.R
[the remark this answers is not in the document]
LiamDBailey: I think throwing an error would be better so that we can detect it if it happens. Could occur due to a bug or be an issue from the data. Either way would be good to know.

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+    dplyr::filter(RingAge == "chick") %>%
+
+    ## Summarize brood information for each nest
+    dplyr::group_by(.data$BreedingSeason, .data$PopID, .data$LocationID) %>%
+
+    dplyr::summarise(Species = names(which.max(table(.data$Species, useNA = "always"))),
+                     FemaleID = names(which.max(table(.data$MotherRing, useNA = "always"))),
+                     MaleID = names(which.max(table(.data$FatherRing, useNA = "always"))),
+                     AvgChickMass = round(mean(Mass[ChickAge <= 16L & ChickAge >= 14L], na.rm = TRUE),1),
+                     NumberChicksMass = sum(ChickAge <= 16L & ChickAge >= 14L & is.na(Mass) == F),
```
LiamDBailey, line 318:
```suggestion
                     NumberChicksMass = sum(ChickAge <= 16L & ChickAge >= 14L & is.na(Mass) == FALSE),
```

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+                     AvgTarsus = round(mean(Tarsus[ChickAge <= 16L & ChickAge >= 14L ], na.rm = TRUE),1),
+                     NumberChicksTarsus = sum(ChickAge <= 16L & ChickAge >= 14L & is.na(Tarsus) == F)) %>%
```
LiamDBailey, line 320:
```suggestion
                     NumberChicksTarsus = sum(ChickAge <= 16L & ChickAge >= 14L & is.na(Tarsus) == FALSE)) %>%
```

tests/testthat/test_GLA_pipeline.R
```diff
@@ -0,0 +1,238 @@
+context("Run data quality check on Glasgow, Scotland, pipeline output")
```
LiamDBailey, line 1: All passing for me now. Once this is ready we'll merge in the master and adjust the setup so it works with the new testing architecture.

LiamDBailey (changes requested):
Looking good. Mostly just comments to make the code a bit cleaner/faster.

One outstanding issue may be how we deal with multiple clutches in the same box for one season. At multiple times (e.g. l. 312 and l. 577) we are joining or summarising by BreedingSeason, PopID and LocationID but this could include multiple clutches from the same year. Even though they don't have 2nd clutches in the population, it could still be possible for there to be multiple clutches with hatched chicks (one might have failed before fledging). How we deal with this is a bit tricky. Perhaps using `fuzzyjoin` to match captures to a particular BroodID?

R/calc_clutchtype.R
```diff
@@ -23,393 +25,798 @@
...
-  if ("LayDate_observed" %in% colnames(data)) {
+  if (protocol_version == "1.0"){
 
-    cutoff_dat <- data %>%
-      dplyr::group_by(PopID, BreedingSeason, Species) %>%
-      dplyr::mutate(cutoff = tryCatch(expr = min(.data$LayDate_observed, na.rm = TRUE) + lubridate::days(30),
-                                      warning = function(...) return(NA))) %>%
-      # Determine brood type for each nest based on female ID
-      dplyr::group_by(BreedingSeason, Species, FemaleID)
+    if ("LayDate_observed" %in% colnames(data)) {
```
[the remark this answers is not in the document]
cwtyson: This is true. The function was actually already written to handle both the new and old protocol versions, but did so by looking for 'LayDate_ovserved' in the column names and processing the data accordingly. I've changed the function so that it now uses the column names based on the version argument. This required changing column names in the pipelines that were formatted to version 1.1 since all of these pipelines first put the column names in the old format and then renamed the columns to the new format before exporting...

R/create_data_table_templates.R
```diff
@@ -0,0 +1,105 @@
...
+                                      FemaleID = NA_character_,
+                                      MaleID = NA_character_,
+                                      ClutchType_observed = NA_character_,
+                                      ClutchType_calculated = NA_character_,
+                                      LayDate_observed = as.Date(NA),
+                                      LayDate_min = as.Date(NA),
+                                      LayDate_max = as.Date(NA),
+                                      ClutchSize_observed = NA_integer_,
+                                      ClutchSize_min = NA_integer_,
+                                      ClutchSize_max = NA_integer_,
```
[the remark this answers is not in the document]
cwtyson: True. Do you think it would be better to change the column class here (even though it would then differ from the standard protocol) or keep it as NA_integer_?

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+#'This section provides details on data management choices that are unique to
+#'this data. For a general description of the standard format please see
+#'\href{https://github.com/SPI-Birds/documentation/blob/master/standard_protocol/SPI_Birds_Protocol_v1.1.0.pdf}{here}.
+#'
+#'\strong{Species}: Primarily great tits and blue tits.
+#'
+#'\strong{IndvID}: Should be a 7 digit alphanumeric string. IndvIDs with more or less characters are likely errors.
+#'
+#'\strong{CaptureDate}: Some individuals were not recorded in the ringing records, but were observed breeding at a monitored nest.
+#'For these individuals, the CaptureDate is set as May 15 of the breeding year.
```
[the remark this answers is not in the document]
cwtyson: Yea, a few birds are only in the nest data. The data owner suggested May 15 as a default date.

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+  start_time <- Sys.time()
+
+  ## Set options
+  options(dplyr.summarise.inform = FALSE)
+
+  ## Read experiment classification table
+  expID_tab <- utils::read.csv(file = paste0(db, "/GLA_experiment_groups.csv"))
+
+  ## Read in primary data from brood records
+  nest_data <- readxl::read_xlsx(path = paste0(db, "/GLA_PrimaryData_Nest.xlsx"), guess = 5000) %>%
```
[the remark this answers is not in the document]
cwtyson: Nice catch, thanks

R/format_GLA.R
cwtyson: Indeed. It is a pain and this chunk to the process the dates is the slowest part of the code. 

Your suggestions make it look much cleaner though.

R/format_GLA.R
```diff
@@ -0,0 +1,688 @@
...
+#'
+#'A pipeline to produce the standard format for the nest box population in Glasgow, Scotland, administered by Davide Dominoni.
+#'
+#'This section provides details on data management choices that are unique to
+#'this data. For a general description of the standard format please see
+#'\href{https://github.com/SPI-Birds/documentation/blob/master/standard_protocol/SPI_Birds_Protocol_v1.1.0.pdf}{here}.
+#'
+#'\strong{Species}: Primarily great tits and blue tits.
+#'
+#'\strong{IndvID}: Should be a 7 digit alphanumeric string. IndvIDs with more or less characters are likely errors.
```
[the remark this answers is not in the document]
cwtyson: Good idea. I've set non-conforming IDs to NA in the brood data and the capture data.

PR #179, opened 2021-06-29
Update tests

LiamDBailey:
This PR should make some major changes to how we run tests.

## Isolate tests
Each pipeline is run individually within the corresponding test, rather than using `run_pipelines()`. This has some downsides but a **major** upside is that we can now isolate and test pipelines individually.

```
#This will run ONLY the test for the PEW pipeline
devtools::test(filter = "PEW")
```

## Change test file names
To allow this filtering process, test files should be named `test-XXX.R` (notice the hyphen rather than underscore). Anything after the hyphen can then be passed to the `filter` argument in `devtools::test()`

## Skip test with wrong OS
Instead of writing if/else statements to check OS, we now skip tests that can't run on Mac using `testthat::skip_on_os("mac")`.

## Pipeline warnings are now a problem
Previously when we ran everything together with `run_pipelines()` in the file `setup-xyz.R` pipelines could throw warnings (e.g. NA coercion warnings) and these were hidden/not a problem. Now each pipeline is run in the test and it will flag *all* these warnings. We need to make sure that pipelines do not return superfluous warnings or that warnings can be turned off by default (e.g. in cases where warnings might help flag weird records).


cwtyson (changes requested):
All looks good to me and I was able to do the tests on my machine.

The choose_directory function just needs to be changed and then it should be good to go.

PR #182, opened 2021-06-30
WRS pipeline

cwtyson:
Work in progress, still many TODO notes that need to be fixed after a chat with the data owner.

Also, the last test location data test fails due to nests that are not occupied by 'legitimate' species being filtered out.


R/format_WRS.R
```diff
@@ -0,0 +1,648 @@
...
+    ## Recode columns
+    ## TODO: Check about age codes
+    ## TODO: Check about aggression scoring - Should this be an experiment? Currently listed as OTHER
+    ## TODO: where() not namespaced
+    ## Check species codes
+    mutate(dplyr::across(where(is.character), ~dplyr::na_if(., "NA")),
+           PopID = "WRS",
+           BreedingSeason = as.integer(.data$BreedingSeason),
+           dplyr::across(c(.data$Mass, .data$WingLength, .data$Tarsus), ~ suppressWarnings(as.numeric(.x))),
+           CaptureTime = suppressWarnings(format(openxlsx::convertToDateTime(.data$Hour), "%H:%M")),
```
LiamDBailey, line 215: This can be done with `janitor::convert_to_datetime()`. We should use this instead of introducing another dependency. I noticed that there are some times stored as numeric and some that are still strings e.g. row 841, so this will probably need some case_when statement to ensure these times aren't coerced to NA.

LiamDBailey (commented):
Few small changes, didn't look in more detail at the moment because it's still a work in progress.

R/format_WRS.R
cwtyson: I couldn't get convert_to_datetime to work, which is why I'd gone with the openxlsx function. I was able to find a base R solution that works though: `format(as.POSIXct(Sys.Date() + as.numeric(adult_data$Hour)), "%H:%M", tz="UTC"))`

This is used conditionally based on the format of the time in order to handle the fact that there is a mix of decimal days and character strings.

LiamDBailey (commented):
I notice there's still a number of outstanding questions here, so this may need to wait till you can talk with the data owner before merging. Left a couple extra comments that might be helpful.

PR #183, opened 2021-07-01
Edit run_pipelines

cwtyson:
I made a few small changes to `run_pipelines` 

The main types of changes were:
- Added `dplyr::ungroup` within each chunk that combines the data frames from each pipeline. This is helpful since if the input data has some grouping structure, it will create duplicate row numbers and this isn't helpful for the quality checks.
- Add namespaces where missing.
- Changed some of the linebreaks 
- Added a missing underscore to the beginning of the 'Individual_data' output filename

Nothing of major importance, but hopefully still helpful.


cwtyson:
I didn't realize that this wasn't branched from the master, so there are many more commits here than I intended to merge...

LiamDBailey (approved):
Can't see any issues here. I've merged the master and deleted the WRS code (we're reviewing that in another PR) and then I'll merge this PR in.

PR #187, opened 2021-08-02
General tests

cwtyson:
This pull request adds functions to run general tests in `test_general_format.R` and implements these general tests in the `test-PEW.R.` This pull request also adds tests for the general tests in `test-general_tests.R`. 

Currently, the PEW pipeline fails several of the tests. All of the failing tests are currently commented out. The tests fail for relatively minor reasons that should be easy to fix. Here is a list of the tests that fail:
- The column class test fails because clutch/brood max classes are integers, which agrees with the standard protocol, but does not allow for Inf.
- ID format tests fail because individual ID's are given as 'unringed' when unknown and this does not match the expected ID format which is only numeric. The unringed individuals can be set to NA or the ID format can be adjusted to also expected 'unringed'
- BroodID has duplicated values, which according to the standard protocol it should not.


cwtyson:
I realized that the updated data templates created problems for selecting/adding columns within format_GLA, so I fixed this bug within the pipeline and confirmed that the branch is passing all checks.

PR #197, opened 2022-12-04
add "TEST" to the first line

ameliefar:
(from my personal computer)

### DIFF
--- a/R/calc_age.R
+++ b/R/calc_age.R
@@ -1,4 +1,4 @@
-#' Calculate age based on when an individual was first captured
+#' Calculate age based on when an individual was first captured - TEST
 #'
 #' Arrange data by individual and year and then determine potential age in each
 #' capture.


LiamDBailey:
@ameliefar Looks to be working well!! Next, try requesting a review from Stefan or me (in the top right).

ameliefar:
Sorry for all of this, I'm trying to figure out why I can't click on "reviewers"

LiamDBailey:
This is what it looks like to me (below). It might be because you don't have admin privileges here. In worst case scenario, we can assign the reviewers instead, although that's not ideal!

space

LiamDBailey:
If that definitely doesn't work for you, we can try adding you with some limited privileges that allow you to manage pull requests. Let me know if you have some success!

StefanVriend:
This is what I find in relation to requesting a pull request review in the Github documentation (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review):

image

LiamDBailey:
Ok, so seems that at this stage @ameliefar wouldn't be able to assign (or be assigned) a reviewer. Is this a problem? Perhaps not, because Stefan or I can assign reviewers once we know the PR exists (if you send a message or tag us in a comment). We can, of course, give @ameliefar write access, but if we want to make this system more general (also to people that we might not know) we don't want a system where each person is expected to get write access.

What do you think @StefanVriend?

StefanVriend:
In light of making things more general (and secure), I think your first suggestion is the best option. I guess it will be rather undemanding, as it shouldn't require more than a few clicks for us to assign a reviewer.

ameliefar:
@LiamDBailey @StefanVriend - I was just checking I could also do the process from my lab computer.

StefanVriend (commented):
Just assigned myself as reviewer, did the review, and now it should be ready to merge.

StefanVriend (approved):
I realise now that I should have clicked 'approve' rather than 'comment'.

PR #209, opened 2024-10-24
Replace odbc driver to access Microsoft Access data by rJava/rJackcess

StefanVriend:
New function `extract_access_db()` converts the relevant tables from an Access database to csv. The pipelines then use the csv files as inputs instead of a connection to the Access db. 

This change affects pipelines `format_NIOO()` and `format_AMM()`.


R/format_WYT.R
jbburant, line 1: did this filter (line 349) break without .data, or is it added for specificity?

R/format_SSQ.R
jbburant, line 1: did this filter (line 349) break without .data, or is it added for specificity?

R/format_MON.R
jbburant, line 1: Use of double backslashes in file path formation is inconsistent with other scripts (e.g. singles used in format_AMM.R), but does not affect functionality.

jbburant (approved):
changes in 20 files reviewed.

PR #241, opened 2025-08-29
Closes #229: Fixes to GLA pipeline

### LINKED ISSUE #229: Fixes to GLA pipeline
While preparing the GLA pipeline [v1.1.0] for testing fairchive and the analytical module, the following issues need solving:

- [x] wrong name of primary data file "GLA_experiment_groups.csv"
  - files that are necessary for pipelines *must* start with "PrimaryData" on Fairchive (e.g., "PrimaryData.xlsx" or "PrimaryData_broods.csv"), *without* the pipeline prefix. The analytical module adds the pipeline prefix (e.g., GLA) before the pipelines are run. The pipeline *must* then call the file *with* the pipeline prefix (e.g., "GLA_PrimaryData_Nest.xlsx"), as we have always done.

### LINKED ISSUE #230: Fix issue in `check_coordinates()` when all records have identical coordinates
`check_coordinates()` calculates a centre point for all unique coordinates per population/site in `Location data`.  There need to be at least 2 unique records for a centre point to be calculated. In some populations/sites, all locations are set to the same coordinate pair, if no more specific coordinates are known. This currently results in an error. 

Solution: skip check if `Location_data %>% dplyr::distinct(.data$Latitude, .data$Longitude) %>% nrow() == 1`

StefanVriend:
and additionally fixes a bug that still lingered in `check_coordinates()` (despite the previous commit https://github.com/SPI-Birds/pipelines/commit/53a858533e8b10a56d41f70190347b557b1e86a6 that closed #230).

### DIFF
--- a/R/format_GLA.R
+++ b/R/format_GLA.R
@@ -67,7 +67,7 @@ format_GLA <- function(db = choose_directory(),
   on.exit(options(original_options), add = TRUE, after = FALSE)
 
   ## Read experiment classification table
-  expID_tab <- utils::read.csv(file = paste0(db, "/GLA_experiment_groups.csv"))
+  expID_tab <- utils::read.csv(file = paste0(db, "/GLA_PrimaryData_ExperimentGroups.csv"))
 
   ## Read in primary data from brood records
   nest_data <- readxl::read_xlsx(path = paste0(db, "/GLA_PrimaryData_Nest.xlsx"),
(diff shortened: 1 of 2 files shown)


r-vd-woude (approved):

r-vd-woude:
I'm not informed enough about this dataset, but the calculation of the centre point looks to be correct (that is the biggest change here).

StefanVriend:
> I'm not informed enough about this dataset, but the calculation of the centre point looks to be correct (that is the biggest change here).

Yes, understandable. The centre point calculation is actually a general piece of code that can be applied to any pipeline, not only GLA, so technically it would have been neater if I had included that into a different PR.

PR #263, opened 2025-10-28
Accessparser python fix

r-vd-woude:
Replacing the rjackcess solution to retrieve data from MS Access DB with a the Python library [AccessParser ](https://pypi.org/project/access-parser/).

Also updated the readme and removed references to Java there.


StefanVriend (changes requested):
@r-vd-woude looking good! Committed a few smaller changes.
Only thing that I wonder: should we remove `test_access_db_amm_remove.R` and `test_access_db_nioo_remove.R` as the functions in those tests are no longer there (e.g., `export_access_db_python()` or `export_access_db_rjackcess()`)?

PR #276, opened 2025-11-13
Fixes to NIOO pipeline

### LINKED ISSUE #234: Fixes to NIOO pipeline
While preparing the NIOO pipeline [v1.1.0] for testing fairchive and the analytical module, the following issues need solving:

- [x] implement general tests (`R/test_general_format.R`) in `tests\testthat\test-NIOO.R`
- [x] implement format validation for `FemaleID`, `MaleID`, `IndvID`
- [x] fix `ExperimentID` in Brood data
- [x] ensure that `LocationID` is unique across sites (also see explanation in #244)

r-vd-woude:
The general tests are not yet implemented, not sure how to proceed on that.
Rest should be fixed

Resolves #234 partly.


StefanVriend (changes requested):
Fixed a bug (https://github.com/SPI-Birds/pipelines/pull/276/commits/b642f444403b4a03be0928b5a5b3398fc56fdd4f), but some of the tests fail in my case. Maybe these are related to me using an older version of the NIOO access db? Will check again next week.

r-vd-woude:
@StefanVriend What is the status on this PR?
Do I still need to make some changes?

StefanVriend:
Yes, @r-vd-woude, maybe I should have made that clearer in my previous comment. Sorry for that. 
When running the tests, a bunch of them failed. Can you run the tests and check if they fail or not?

r-vd-woude:
### LocationType:

The NIOO database has some nestbox types in their NestBoxType table that are hard to classify.
NestBoxTypes that we could no easily classify, have been assigned as mistnet captures for now.
We might need to update that later, but, they are about half of the 7000 records in the Location_data table.

### BroodSize_max and BroodSize_min:

As the database sometimes has NA (or empty cells) for NumberHatchedDeviation, simply adding that together with NumberHatched will always produce NA's. The same goes for the other way around.
Hence, we introduce some rules, that when one of the two is NA, that becomes a zero, so adding them together still works.
When both are NA, it will become and stay NA.
See lines 376 to 455.

r-vd-woude:
@StefanVriend 

Now passes all tests, see above comment for choices I made regarding some columns.

StefanVriend (changes requested):
13 tests are not passing in my case. For example, the brood data test fails because of `ExperimentID` is expected to be 'SAMUEL' for `BroodID` 64597 (see line 133 in 'test-NIOO.R'. That it fails is good, because that should not be a possible value.

In the files changed tab of this PR, I don't see any changes made to 'test-NIOO.R'. @r-vd-woude, did you push all commits?

r-vd-woude:
> 13 tests are not passing in my case. For example, the brood data test fails because of `ExperimentID` is expected to be 'SAMUEL' for `BroodID` 64597 (see line 133 in 'test-NIOO.R'. That it fails is good, because that should not be a possible value.
> 
> In the files changed tab of this PR, I don't see any changes made to 'test-NIOO.R'. @r-vd-woude, did you push all commits?

For my understanding: the current test does not follow the protocol then?
So, I should update the test to it correctly follows the protocol?

r-vd-woude:
Should be good?

image

StefanVriend:
Part of the tests in the 'test-XXX.R' file are intended to make sure that the content of the raw data files are the same as the pipeline outputs. That is, that we do not wrongly change values where they shouldn't have been changed. Indeed, as my example above shows, in the first version of the tests we were not so strict with following the protocol.

Often when updating a pipeline, updating the content tests are also necessary, whereas the tests that check for form and format (so the tests in 'test_general_format.R') should ideally be static. 

I think it's worth discussing these content checks a bit more; maybe they are a bit non-sensical, if we apparently have to change them every time we update the pipeline?

StefanVriend:
> Should be good?
> 
> image

Ahhh, I see, my local version of this branch didn't sync with origin... my bad, sorry.

Edit: that might also be the reason why I failed to run all tests successfully for WRS (see #182)

Edit2: that aside, I think it might still be worth discussing the content tests that I refer to in my previous comment.

StefanVriend:
> Should be good?
> 
> image

@r-vd-woude which command do you use to run these tests?

tests/testthat/test-NIOO.R
```diff
@@ -56,202 +55,200 @@ test_that("Individual data returns an expected outcome...", {
...
+  # All records have expected LocationType
+  expect_true(all(subset(NIOO_data, LocationID == "OOS_1841")$LocationType == "MN"))
+  # Expect Start and EndSeason of first box at this location is as expected
+  expect_equal(subset(NIOO_data, LocationID == "OOS_1841")$StartSeason, 1977L)
+  expect_equal(subset(NIOO_data, LocationID == "OOS_1841")$EndSeason, NA_integer_)
+  # Check that LocationID is in the expected PopID
+  expect_equal(subset(NIOO_data, LocationID == "OOS_1841")$PopID, "OOS")
 })
 
 ## General tests
```
StefanVriend, line 248: I think it would be preferred to use your new wrapper `test_protocol_compliance(pipeline_output)` here?

StefanVriend (changes requested):

StefanVriend (approved):

PR #278, opened 2025-11-14
Removed duplicate individuals in Individual_data in format_GLA

r-vd-woude:
Implemented a fix to remove the duplicated invididuals in Individual_data.

We might need to ask the data owner if this is correct.


StefanVriend:
Did you by any chance have a look at what the reason was for the duplication? 

As GLA is a multi-site pipeline (5 or 6 locations in a rural-urban gradient extending from Glasgow), could it be simply be the same individual seen at multiple locations?

r-vd-woude:
@StefanVriend good call, I did not.
For HOC, the individuals were just duplicated rows, containing the exact same info.
I assumed here it would be the same, but I will check it to be sure.

StefanVriend:
Conclusion from meeting with @r-vd-woude:

Removing duplicates by keeping the record with most complete information. If 2 or more records are equally complete, keep the record with the PopID it was first captured in.

StefanVriend (approved):

PR #281, opened 2025-11-20
Added extra tests and wrapper function for protocol tests from #274

### LINKED ISSUE #274: Add other protocol validation checks to tests
The protocol validation checks that the ICM system executes whenever a pipeline output is ingested often result in similar errors because we do not check these yet as part of the testthat tests.

Tests that are currently missing:

- [ ] Chick age >= 0
- [ ] Maximum brood size/clutch size/fledgling number >= 1

Some of these might actually be covered as part of the quality checks, so make sure to check that first. Also see #228.

r-vd-woude:
Quite a big pull request.
Resolves #274.

Did some testing, seems all to work.


R/test_protocol_wrapper.R
```diff
@@ -0,0 +1,127 @@
...
+#' @param protocol_version The protocol version of the SPI-Birds standard
+#' data being used. Either "1.0.0", "1.1.0", or "2.0.0".
+#' @param ID_format Regular expression pattern for validating ID formats.
+#' Default is FALSE, which skips this test.
+#'
+#' @return Invisible TRUE if all tests pass, errors otherwise.
+#' @export
+
+test_protocol_compliance <- function(pipeline_output,
+                                     protocol_version,
```
StefanVriend: As the protocol version is part of pipeline output (as the fifth item in the list), does it need to be a separate argument here?

StefanVriend (approved):
Great! This works super for now.
But in 2026, we will need to change how the tests are set up, as the data will no longer be hosted on our local server but on the Dataverse repo. Then we would need to refer to the DOI, I guess? I will create a issue for it.

PR #326, opened 2026-06-11
Added locations, and species - locations combinations to the pop*.csv files

r-vd-woude:
Updated some metadata files.
Allows us to run pipelines only for specific species.


StefanVriend (changes requested):
I suggest assigning these as part of the respective PRs themselves, so that pop_species_codes and pop_locations in the master branch only contain those entries that have a functional and merged pipeline

PR #327, opened 2026-07-15
Created MAY 1.1 pipeline

### LINKED ISSUE #322: Create 1.1 pipeline for MAY
## Details
Name: Mayachino
ID: MAY
Data last shared: July 2022
Data format: Excel/csv

Follow standard protocol version: 1.1.0

## Existing code to build from
Stale branch: https://github.com/SPI-Birds/pipelines/blob/MAY/R/format_MAY.R
**Note**: this branch is based on an outdated version most similar to 2.0.0, so not sure how helpful this is

r-vd-woude:
Created a complete MAY pipeline.

Passes all tests.

Contacted the data owner already with a few questions, but those are small.


StefanVriend (commented):
Impressive work, especially with the parsing of all the different IDs.

What's the status of the TODOs and the data custodians answers on them? Do you judge any of them "essential" before we merge this with the master branch?

r-vd-woude:
@StefanVriend 

The only thing left, is that some ring sequences extend beyond 15 rings (I made a cut-off at 15, that is arbitrary).
Feedback from data-owner is incorporated, only outstanding question currently is the 15+ ring sequences.

I don't think that is essential for merging and can be fixed easily once we have an answer.

StefanVriend (approved):
Good to go!
@r-vd-woude: do I understand correctly that the data owner was okay with the 15+ sequences being solved later?

r-vd-woude:
> Good to go! @r-vd-woude: do I understand correctly that the data owner was okay with the 15+ sequences being solved later?

@StefanVriend. Those issues are now gone: it was a data problem, not a pipeline one.

StefanVriend:
Oh right, that's what you said about the 20+ records that the data owner fixed. Thanks!

